PR #1, opened 2023-11-08
feat: Introduce macros for generating dispatching func, inputs and outputs

DennisInSky:
Introduces `command_handlers` and `query_handlers` attributes for modules implementing some handlers. Any standalone function of the module regardless of the function visibility  is considered to be a handler. Based on the functions signature, the macro generates 2 enums: one for input params, another for output ones, and a function which takes an input enum variant as a param, dispatches it to one of the defined handlers, and returns an output enum variant as a result with indication whether the call was successful or not.


examples/this-that/wasm/src/lib.rs
```diff
@@ -0,0 +1,30 @@
...
+    let command = msg::load::<commands::Commands>()
+        .expect("This needs to be handled in a consistent way: input parse error");
+    let (result, is_error) = commands::handlers::handle_commands(command);
+    let result = result.encode();
+    if is_error {
+        unsafe {
+            gsys::gr_panic(result.as_ptr(), result.len() as u32);
+        }
+    }
+    msg::reply(result, 0).expect("This needs to be handled in a consistent way: reply error");
```
NikVolf, line 19: why 0 gas ?

NikVolf:
Async example right away would be nice

examples/this-that/app/src/lib.rs
```diff
@@ -0,0 +1,93 @@
...
+        p4: TupleStruct,
+    ) -> Result<(String, u32), String> {
+        debug!("Handling 'do_this': {}, {}, {:?}, {:?}", p1, p2, p3, p4);
+        Ok((p2, p1))
+    }
+
+    // That
+    fn do_that(
+        param: DoThatParam,
+    ) -> Result<(String, u32), (String,) /* IDEA can't handle meta with single element tuple */>
```
NikVolf, line 59: why do you need here single element tuple anyway?

examples/this-that/wasm/src/lib.rs
DennisInSky: This is not gas, but value. As far as I get transferring value from a contract is not used

examples/this-that/app/src/lib.rs
DennisInSky: This is just as many possible type usages as I could imagine.

DennisInSky:
> Async example right away would be nice

Do we want to generate async version of `handle_commands` all the time regardless of whether there are async handlers or not?

examples/this-that/wasm/src/lib.rs
```diff
@@ -0,0 +1,30 @@
...
+use gstd::msg;
+use parity_scale_codec::Encode;
+use this_that_app::{commands, queries};
+
+// TODO: This function might be generated by some macro like `gprogram`
+#[no_mangle]
+extern "C" fn handle() {
+    let command = msg::load::<commands::Commands>()
+        .expect("This needs to be handled in a consistent way: input parse error");
+    let (result, is_error) = commands::handlers::handle_commands(command);
```
shamilsan, line 12: It is more idiomatic not to include a module for type and include only one level to function (https://doc.rust-lang.org/book/ch07-04-bringing-paths-into-scope-with-the-use-keyword.html#creating-idiomatic-use-paths).
```suggestion
    let command: Commands = msg::load()
        .expect("This needs to be handled in a consistent way: input parse error");
    let (result, is_error) = handlers::handle_commands(command);
```

macros/core/src/lib.rs
```diff
@@ -0,0 +1,139 @@
...
+
+                fn do_this(p: SomeStruct) {}
+            }
+        };
+        let expected = quote! {
+            mod commands {
+                extern crate parity_scale_codec as commands_scale_codec;
+                extern crate scale_info as commands_scale_info;
+
+                #[derive(commands_scale_codec::Encode, commands_scale_codec::Decode, commands_scale_info::TypeInfo)]
```
shamilsan, line 59: Too wordy too
```suggestion
                #[derive(Encode, Decode, TypeInfo)]
```

Cargo.toml
```diff
@@ -0,0 +1,29 @@
...
+members = [
+    "examples/this-that/app",
+    "examples/this-that/wasm",
+    "macros",
+    "macros/core",
+]
+
+[workspace.dependencies]
+convert-case = { package = "convert_case", version = "0.6" }
+gmeta = { git = "https://github.com/gear-tech/gear.git", tag = "v0.3.3" }
```
shamilsan, line 18: You can update deps to `v1.0.1`

shamilsan (approved):
It looks good to me as a starting point.

macros/core/src/lib.rs
DennisInSky: I've done it on purpose for the generated code in order to avoid possible name conflicts

DennisInSky:
@NikVolf , @breathx any feedback, approves from you guys?

DennisInSky:
It will be merged in via a child branch

PR #2, opened 2023-11-14
feat: Introduce basic idlgen

DennisInSky:
A library generating IDL file in a format similar to the DFinity's Candid one.


idlgen/src/type_names.rs
```diff
@@ -0,0 +1,188 @@
...
+fn primitive_type_name(type_def: &TypeDefPrimitive) -> Result<String> {
+    match type_def {
+        TypeDefPrimitive::Bool => Ok("bool".into()),
+        TypeDefPrimitive::Char => Ok("char".into()), // Candid doesn't have it. Do we want to support it? If such it will require a definition
+        TypeDefPrimitive::Str => Ok("text".into()),
+        TypeDefPrimitive::U8 => Ok("nat8".into()),
+        TypeDefPrimitive::U16 => Ok("nat16".into()),
+        TypeDefPrimitive::U32 => Ok("nat32".into()),
+        TypeDefPrimitive::U64 => Ok("nat64".into()),
+        TypeDefPrimitive::U128 => Ok("nat128".into()), // Candid doesn't have it
```
holykol, line 179: Could be `[u8; 16]`, but candid doesn't have sized arrays either. Perhaps make it `vec nat8`? Not a blocker though

DennisInSky: I think this is more a question of how to represent it on a client side. Probably, that's the reason why Candid doesn't support it, and I'm leaning towards throwing it away for us too.

DennisInSky: Sized arrays are pretty much a question of the same matter - not all languages have a concept of fixed size arrays where the size is known at compile time.

DennisInSky:
It will be merged in via a child branch

PR #4, opened 2023-12-03
implement Rust client

holykol:
goals (open to discussion):
* enable easier testing and invocations from Rust

non-goals:
* reinvent all of the options of sending a mesasge (i.e. [send_delayed](https://docs.gear.rs/gstd/msg/fn.send_delayed.html#), [send_from_reservation](https://docs.gear.rs/gstd/msg/fn.send_from_reservation.html), [send_bytes_with_gas](https://docs.gear.rs/gstd/msg/fn.send_bytes_with_gas.html))


DennisInSky:
The thing is that I'm re-building the entire server side code generation which means this client will require re-implementation. As we discussed it has to be built using idl which remains unchanged even I after I reimplement the way server side is generated

holykol:
changes will be ported to another PR

PR #15, opened 2023-12-12
Change idl format

osipov-mit:
- Rename `record` -> `struct`
- Change fields separator to `,`
- Add semicolon after function declarations
- Rename `variant` -> `enum`

### DIFF
--- a/idlgen/hbs/composite.hbs
+++ b/idlgen/hbs/composite.hbs
@@ -1,5 +1,5 @@
-type {{{lookup @root/type_names id}}} = record {
+type {{{lookup @root/type_names id}}} = struct {
 {{#each fields}}
-  {{#if name}}{{name}}: {{/if}}{{{lookup @root/type_names type}}};
+  {{#if name}}{{name}}: {{/if}}{{{lookup @root/type_names type}}},
 {{/each}}
 };
--- a/idlgen/hbs/idl_ex.hbs
+++ b/idlgen/hbs/idl_ex.hbs
@@ -6,9 +6,9 @@
 {{/each}}
 service {
 {{#each commands.type.def.variant.variants}}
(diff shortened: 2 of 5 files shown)


idlgen/hbs/idl_ex.hbs
```diff
@@ -6,9 +6,9 @@
 {{/each}}
 service {
 {{#each commands.type.def.variant.variants}}
-  {{name}} : ({{#with (lookup @root/all_types fields.[0].type)}}{{#each type.def.composite.fields}}{{#if @index}}, {{/if}}{{name}}: {{{lookup @root/type_names type}}}{{/each}}{{/with}}) -> {{{lookup @root/type_names fields.[1].type}}}
+  command {{name}} : ({{#with (lookup @root/all_types fields.[0].type)}}{{#each type.def.composite.fields}}{{#if @index}}, {{/if}}{{name}}: {{{lookup @root/type_names type}}}{{/each}}{{/with}}) -> {{{lookup @root/type_names fields.[1].type}}};
```
holykol, line 9: I would leave this without prefix. Imo, the semantics would be simpler if methods can mutate state by default unless `query` is specified

idlgen/hbs/composite.hbs
```diff
@@ -1,5 +1,5 @@
-type {{{lookup @root/type_names id}}} = record {
+type {{{lookup @root/type_names id}}} = struct {
```
holykol, line 1: We might as well change `variant` to `enum` to keep it consistent with Rust

osipov-mit: agreed

idlgen/hbs/idl_ex.hbs
osipov-mit: make sense

holykol (approved):

PR #16, opened 2023-12-15
feat: Implement Rust client generation

holykol:
Adds generation of client from IDL 

- [x] client generation from JS parser json output 
- [x] use rust parser now that it's out
- [x] NativeSender
- [ ] mockall (will need help with feature flags)
- [ ] naming for all traits
- [ ] tests

<details>
<summary>old todo:</summary>

- [ ] port code macros from old PR to generate client straight from code skipping generator
- [ ] ensure all three types clients work (now we have stable async traits now)
    - [ ] GClientSender
    - [ ] GTestSender
- [ ] ensure all use cases are supported (e.g. rpc -> program, program -> program)
- [ ] more tests
- [ ] docs on generator binary?

</details>


holykol:
still some todo's

client/src/lib.rs
```diff
@@ -0,0 +1,109 @@
...
+#[non_exhaustive]
+pub struct SendTicket<R: Decode + Debug> {
+    // message we are waiting on
+    f: MessageFuture,
+    // silence the compiler
+    _marker: PhantomData<R>,
+}
+
+impl<R: Decode + Debug> SendTicket<R> {
+    pub async fn result(self) -> Result<R, gstd::errors::Error> {
```
DennisInSky, line 24: 1) I would call it `response` in order to avoid any confusion with the `Result` type
2) Rust community suggests using types w/o full path and use aliases if it is needed, for example GStdError

client/src/lib.rs
```diff
@@ -0,0 +1,109 @@
...
+
+#[non_exhaustive]
+#[derive(Default)]
+pub struct SendArgs {
+    value: u128,
+    _sender_id: [u8; 32],
+}
+
+#[non_exhaustive]
+pub struct SendTicket<R: Decode + Debug> {
```
DennisInSky, line 16: Any reason why SendTicket should be parameterized with R instead of its `response` method?

client/src/lib.rs
```diff
@@ -0,0 +1,109 @@
...
+        R::decode(&mut payload.as_ref()).map_err(gstd::errors::Error::Decode)
+    }
+
+    pub fn message_id(&self) -> MessageId {
+        self.f.waiting_reply_to
+    }
+}
+
+/// Sender that runs message against gtest::Program
+#[derive(Default)]
```
DennisInSky, line 36: What's the point in this default?

client/src/lib.rs
```diff
@@ -0,0 +1,109 @@
...
+pub struct NativeSender;
```
DennisInSky, line 37: -> GStdSender
+ messing empty line between struct and impl

client/src/lib.rs
```diff
@@ -0,0 +1,109 @@
...
+impl NativeSender {
+    pub fn new() -> Self {
+        Self
+    }
+}
+
+#[derive(Default)]
```
DennisInSky, line 44: There is no point in default impl

client/src/lib.rs
```diff
@@ -0,0 +1,109 @@
...
+pub struct Call<R: Decode + Debug> {
```
DennisInSky, line 45: Seems like R parameter is not needed it is more for the `response` method being parameterized with expected response type

client/src/lib.rs
```diff
@@ -0,0 +1,109 @@
...
+    /// request payload
+    payload: Vec<u8>,
+    /// optional args
+    args: SendArgs,
```
DennisInSky, line 49: Not used for now

client/src/lib.rs
```diff
@@ -0,0 +1,109 @@
...
+    /// silence the compiler
+    _marker: PhantomData<R>,
+}
+
+#[non_exhaustive]
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum SendError {
+    Parse(parity_scale_codec::Error),
```
DennisInSky, line 57:
```suggestion
    Parser(ParserError),
```
to be consistent with `Sender`

client/src/lib.rs
```diff
@@ -0,0 +1,109 @@
...
+impl<R: Decode + Debug> Call<R> {
+    pub fn new(payload: Vec<u8>) -> Self {
+        Self {
+            payload,
+            args: Default::default(),
+            _marker: PhantomData,
+        }
+    }
+
+    pub async fn send(self, address: impl Into<[u8; 32]>) -> Result<SendTicket<R>, SendError> {
```
DennisInSky, line 82:
```suggestion
    pub async fn send(self, dest: ActorId) -> Result<SendTicket<R>, SendError> {
```

client/src/lib.rs
```diff
@@ -0,0 +1,109 @@
...
+
+impl<R: Encode + Decode + Debug> Call<R> {
+    /// Create call that instantly resolves into `result`. Useful for mocking responses
+    pub fn ready(result: R) -> Self {
+        let mut payload = Vec::new();
+        result.encode_to(&mut payload);
+
+        Self::new(payload)
+    }
+}
```
DennisInSky, line 109: There is no point in this for now

DennisInSky (approved):
Partial review

DennisInSky (changes requested):
Partial review

client/src/lib.rs
holykol: Makes sense to me because we already sent out request and are waiting for certain type of response. Reinterpreting it as some other response would be wrong

client/src/lib.rs
holykol: Clippy made me do it! https://rust-lang.github.io/rust-clippy/master/#/new_without_default

client/src/lib.rs
holykol: I'm afraid moving `R` to method will require type annotations from caller program. If the Call is parametrized like this in generated client, we are guaranteed to parse correct type of response. `R` can't be changed since payload serialized at this point.

client/src/lib.rs
holykol: added `with_value`, `with_reply_deposit`

.gitignore
```diff
@@ -1,10 +1,6 @@
...
-
-.vscode
-target/
 .binpath
+.DS_Store
 .vscode
-#.log
+js/lib/
+node_modules/
+target/
```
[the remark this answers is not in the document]
holykol: thx, found a setting for this

client/src/lib.rs
```diff
@@ -0,0 +1,109 @@
...
+use core::marker::PhantomData;
+use gstd::{msg::MessageFuture, prelude::*, ActorId, MessageId};
+use parity_scale_codec::{Decode, Encode};
+
+#[non_exhaustive]
+#[derive(Default)]
+pub struct SendArgs {
+    value: u128,
+    _sender_id: [u8; 32],
+}
```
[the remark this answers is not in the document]
holykol: added `Call::with_value`

client/src/lib.rs
holykol: renamed

client/src/lib.rs
holykol: done

client/src/lib.rs
```diff
@@ -0,0 +1,109 @@
...
+pub struct Call<R: Decode + Debug> {
+    /// request payload
+    payload: Vec<u8>,
+    /// optional args
+    args: SendArgs,
+    /// silence the compiler
+    _marker: PhantomData<R>,
+}
+
+#[non_exhaustive]
```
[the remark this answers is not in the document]
holykol: SendError is the perfect case for non_exhaustive, especially if someone will soon depend on this code

client/src/lib.rs
holykol: Changed, but my vision here was to abstract all libraries so `Call` is universal across all environments. Using `gstd::ActorId` requires importing `gstd`, which as far as I understand isn't compatible with `std`

idlparser/src/ast/mod.rs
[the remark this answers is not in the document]
holykol: See `args_struct(f: &Func) -> Type` where new type is created out of thin air. I think this problem is not exclusive to my implementation (i.e. visitor/non-visitor). CMIIW but this can only be avoided by writing a lot more unnecessary code and duplicating existing structs.

DennisInSky (changes requested):
Some components have wrong responsibilities. I envision it like the below pseudo-code
```
// We have a service client trait + impl
// The impl responsibility is building a `Call` for each method existing in a service.
// This pretty much what we are doing now. The difference will be in what the `Call` does
trait MyServiceClient {
    type Call : CallTrait;

    fn do_something(p1, p2) -> Call<DoSomethingResp>;
}

struct MyServiceClientImpl<'a> {
    sender: &'a GStdSender // This will be replaced by trait
}

impl MyServiceClient for MyServiceImpl {
    fn do_something(p1, p2) -> Call<DoSomethingResp> {
       // Not more than this - just compose the call instance and pass the sender to it
       // so it can use it for sending the encoded payload + decode the response
        Call::<DoSomethingResp>::new(self.sender, "DoSomething", (p1, p2))
    }
}

// The call responsibility is to be able to encode request and decode response
// for building payload + keeping record of `value` + `deposit` etc
// The call should be an implementation of some `Call` trait this will allow us
// to mock `SomeServiceClient` along with immediate reponse `Call`, i.e. we want
// to bypass using the sender
impl Call<TRequest, TResponse>
where TRequest : Encode,
TResponse : Decode {
    fn new(request : TRequest) {...}

    fn send(destination_id) -> CallTicket<TResponse> {
         let payload = [
		    self.func_name.encode(),
			self.args.encode()
		].concat();
		let future_result = self.sender.send(destination_id, payload, value);
		let future = future_result?; // send error
		return CallTicket(self.func_name, future);
    }
}

impl CallTicket
{
	pub fn response() {
		ler response_bytes = self.future.await?;
		let response_func_name = String::decode(&mut response_bytes)?;
		if response_func_name != self.func_name {
			return error;
		}
		R::decode(...);
	}
}
```

client-gen/src/generator.rs
```diff
@@ -0,0 +1,459 @@
...
+
+    fn visit_type(&mut self, t: &'ast Type) {
+        let mut type_generator = TypeGenerator::new(t.name());
+        type_generator.visit_type(t);
+        self.code.push_str(&type_generator.code);
+    }
+}
+
+#[derive(Default)]
+struct RequestStructsGenerator {
```
DennisInSky, line 77: You won't need these structs at all if you follow the suggested pseudo-code. You can always use tuples
```
let payload = [
    "MyFuncName".encode(),
    (p1, p2, p3, ...).encode()
].concat();
```

holykol: rewritten

examples/puppeteer/wasm/build.rs
```diff
@@ -0,0 +1,35 @@
...
+        .expect("Path should have subdirs in the `target` dir")
+        .as_os_str()
+        .to_string_lossy()
+        .into();
+
+    let mut target_dir = out_dir.to_path_buf();
+    while !target_dir.ends_with("target") && target_dir.pop() {}
+
+    target_dir.join("wasm32-unknown-unknown").join(profile)
+}
```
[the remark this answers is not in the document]
holykol: done

client-gen/src/main.rs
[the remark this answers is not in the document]
holykol: rewritten to use in build.rs

sender/Cargo.toml
```diff
@@ -0,0 +1,10 @@
+[package]
+name = "sails-sender"
+version.workspace = true
+authors.workspace = true
+edition.workspace = true
+license.workspace = true
+
+[dependencies]
+gstd = { workspace = true, features = ["debug"] }
```
DennisInSky, line 9: We don't need debug feature enabled by default

sender/src/lib.rs
```diff
@@ -0,0 +1,131 @@
...
+    /// the client to send the message
+    sender: &'a GStdSender,
+    /// silence the compiler
+    _marker: PhantomData<R>,
+}
+
+#[non_exhaustive]
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum SendError {
+    Parse(ParseError),
```
DennisInSky, line 52:
```suggestion
    Parser(ParseError),
```
So it is consistent with `Sender`

sender/src/lib.rs
```diff
@@ -0,0 +1,131 @@
...
+}
+
+impl From<GStdError> for SendError {
+    fn from(e: GStdError) -> Self {
+        Self::Sender(e)
+    }
+}
+
+impl<'a, R: Decode + Debug> Call<'a, R> {
+    pub fn new<T: Encode + Debug>(method: &str, args: T, client: &'a GStdSender) -> Self {
```
DennisInSky, line 69: I don't think it is correct to have generic method around the request type - it sounds like we can have multiple request types against the same response type which is wrong. We have pairs `RequestType <-> ResponseType` and this is declared by the call

sender/src/lib.rs
```diff
@@ -0,0 +1,131 @@
...
+        let capacity = method.len() + 1 + args.encoded_size();
+        let mut payload = Vec::with_capacity(capacity);
+        payload.extend_from_slice(method.as_bytes());
+        payload.push_byte(b'/');
```
DennisInSky, line 73: Encoding is wrong. It has to be, there are no any slashes
`|encoded-method-name||encoded-payload|`
See example here https://github.com/gear-tech/sails/blob/cc3a4a4f29eb7123216ab3ff6361a96a032482a1/examples/rmrk/resource/wasm/tests/resources.rs#L203

sender/src/lib.rs
DennisInSky, line 69: I would make sender going first

examples/puppeteer/wasm/Cargo.toml
```diff
@@ -0,0 +1,15 @@
...
+[dependencies]
+gstd.workspace = true
+sails-sender.workspace = true
+puppeteer-app = { path = "../app" }
+
+[build-dependencies]
+gwasm-builder.workspace = true
+sails-idlgen.workspace = true
+puppeteer-app = { path = "../app" }
+this-that-svc-app = { path = "../../this-that-svc/app" }
```
DennisInSky, line 15: Doesn't seem needed

examples/this-that-svc/app/src/lib.rs
```diff
@@ -40,6 +40,7 @@ impl MyService {
 }
 
 #[derive(Debug, Decode, TypeInfo)]
+#[allow(unused)]
```
DennisInSky, line 43: Are you sure it is needed? Atm it compiles w/o this

client-gen/Cargo.toml
```diff
@@ -0,0 +1,14 @@
...
+edition.workspace = true
+license.workspace = true
+
+[dependencies]
+anyhow.workspace = true
+convert-case.workspace = true
+sails-idlparser.workspace = true
+
+[dev-dependencies]
+insta = "1.34.0"
```
DennisInSky, line 14: To the workspace deps

client-gen/src/generator.rs
```diff
@@ -0,0 +1,438 @@
...
+        let fn_name = func.name().to_case(Case::Snake);
+
+        self.code
+            .push_str(&format!("fn {}(&{} self,", fn_name, mutability));
+
+        visitor::accept_func(func, self);
+
+        self.code.push_str("{\n");
+
+        let method_name = func.name().to_case(Case::Pascal);
```
DennisInSky, line 174: There is no need to do any conversion here as func name comes exactly in the expected case

DennisInSky (changes requested):

DennisInSky (approved):

PR #33, opened 2024-01-04
feat(idl-parser): Intoduce IDL parser written in Rust

DennisInSky:
Introduces basics for IDL parser written in Rust with an ability of being compiled into WASM.
There are still some issues:
- Can't handle IDL types generated from Rust generics (will be resolved after [this PR](https://github.com/gear-tech/sails/pull/35)
- ~Allows more than 1 unnamed service~ ([commit](https://github.com/gear-tech/sails/pull/33/commits/b897bc06adf202273bb984212a5b80a619e0fc81))
- ~Missing proper FFI~ ([commit](https://github.com/gear-tech/sails/pull/33/commits/5fdd145f408154fb1341bc7fa5ffeef0405224c2))
- ~Missing Visitor for traversing through produced AST~ ([commit](https://github.com/gear-tech/sails/pull/33/commits/901a5c8e043a18455aa3ff7142a10aeaaa351ac4))


idlgen/src/lib.rs
```diff
@@ -340,6 +248,9 @@ service {
   query That : (pr1: SailsIdlgenTestsThatParam) -> str;
 }
 ";
+
         assert_eq!(generated_idl, EXPECTED_IDL);
+        // assert!(generated_idl_program.is_ok());
+        // assert_eq!(generated_idl_program.unwrap().items.len(), 8);
```
shamilsan, line 254: Dead code?

idlparser/src/ffi/types/mod.rs
```diff
@@ -0,0 +1,77 @@
+use crate::types;
+use std::{slice, str};
+
+pub mod visitor;
+
+/// # Safety
+///
+/// See the safity documentation of [`slice::from_raw_parts`].
```
shamilsan, line 8:
```suggestion
/// See the safety documentation of [`slice::from_raw_parts`].
```

idlparser/src/ffi/types/mod.rs
```diff
@@ -0,0 +1,77 @@
...
+#[no_mangle]
+pub unsafe extern "C" fn parse_idl(idl_ptr: *const u8, idl_len: u32) -> *mut Program {
```
shamilsan, line 10: There is no need to make it `pub` and `unsafe` generally if you use it only for FFI. Or do you want to underline that it contains unsafe code?

```suggestion
extern "C" fn parse_idl(idl_ptr: *const u8, idl_len: u32) -> *mut Program {
```

idlparser/src/ffi/types/mod.rs
```diff
@@ -0,0 +1,77 @@
...
+pub unsafe extern "C" fn free_program(program: *mut Program) {
+    if program.is_null() {
+        return;
+    }
+    unsafe {
+        drop(Box::from_raw(program));
+    }
+}
+
+pub type Program = types::Program;
```
shamilsan, line 31: Why type alias instead of `use crate::types::Program`?

idlparser/src/ffi/types/mod.rs
```diff
@@ -0,0 +1,77 @@
...
+
+pub type Service = types::Service;
+
+#[repr(C, packed)]
```
shamilsan, line 35: Why `packed`? Will it be compatible with FFI?

idlparser/src/ffi/types/visitor.rs
```diff
@@ -0,0 +1,295 @@
...
+    let enum_variant = unsafe { enum_variant.as_ref() }.unwrap();
+    let visitor = unsafe { visitor.as_mut() }.unwrap();
+    let mut visitor = VisitorWrapper(visitor);
+    raw_visitor::accept_enum_variant(
+        unsafe { enum_variant.raw_enum_variant.as_ref() }.unwrap(),
+        &mut visitor,
+    );
+}
+
+macro_rules! fn_ptr_addr {
```
shamilsan, line 151: May we replace this macro with a function?

idlparser/src/types/mod.rs
```diff
@@ -0,0 +1,374 @@
...
+          };
+  
+          type T = enum { One }
+        ";
+
+        let program = parse_idl(program_idl).unwrap();
+
+        assert_eq!(program.types().len(), 4);
+
+        //println!("ast: {:#?}", program);
```
shamilsan, line 301:
```suggestion
```

idlparser/src/types/mod.rs
```diff
@@ -0,0 +1,374 @@
...
+impl StructField {
+    pub(crate) fn new(name: Option<String>, r#type: TypeDecl) -> Self {
+        Self { name, r#type }
+    }
+
+    pub fn name(&self) -> Option<&str> {
+        self.name.as_deref()
+    }
+
+    pub fn r#type(&self) -> &TypeDecl {
```
shamilsan, line 221: I'd avoid such names. Maybe `kind` or another synonym?

shamilsan (commented):
Some things are to be explained, maybe we can schedule a call with some kind of presentation?

idlgen/src/lib.rs
DennisInSky: Not quite. It waits for [this PR](https://github.com/gear-tech/sails/pull/35). Going to merge it and uncomment this

idlparser/src/types/mod.rs
holykol: I've used https://crates.io/crates/insta in my PR just to keep track of this output changing. You probably want this: https://docs.rs/insta/latest/insta/macro.assert_debug_snapshot.html

idlparser/src/ffi/types/mod.rs
DennisInSky: Clippy wants it to be marked unsafe as it dereferences raw ptr. `pub` doesn't matter much in this case, but why not

idlparser/src/ffi/types/mod.rs
holykol: agreed, from the [docs](https://doc.rust-lang.org/nomicon/other-reprs.html#reprpacked):

`repr(packed) is not to be used lightly. Unless you have extreme requirements, this should not be used.`

DennisInSky: Don't we use it a lot in our backend? Like here https://github.com/gear-tech/gear/blob/62cd946ed3a851a4f80deda768df7fa213f5171b/gsys/src/lib.rs#L70
Do you think it is better to use `repr(C)` with native padding?

idlparser/src/ffi/types/mod.rs
DennisInSky: Wanted to make it consistent like all FFI types are defined here even though sometimes via aliasing. Is there anything wrong with it?

idlparser/src/types/mod.rs
DennisInSky: This is not for tracking changes in output, but for seeing what it looks like

idlparser/src/ffi/types/mod.rs
DennisInSky: How repr(C) will be achieved in 'JS' @osipov-mit ?

holykol: > https://github.com/gear-tech/gear/blob/62cd946ed3a851a4f80deda768df7fa213f5171b/gsys/src/lib.rs#L70

I think things are `repr(packed)` in backend because of light memory requirements or some ABI compatibility reasons. We should avoid using `packed` here if removing it doesn't break things.

idlparser/src/types/mod.rs
holykol: +1 for kind

holykol (approved):
Would be great if we had some basic example/test using parser from different language so we know it works. 
I guess that can wait because we already have functioning parsers for both JS and Rust.

DennisInSky:
> Ru

We don't need separate parsers for JS and Rust. We need the only one which works everywhere as we want to have a single place of truth, i.e. grammar

idlparser/src/types/mod.rs
```diff
@@ -0,0 +1,374 @@
...
+
+    pub fn type_decl(&self) -> &TypeDecl {
+        &self.type_decl
+    }
+}
+
+#[derive(Debug, PartialEq)]
+pub struct Type {
+    name: String,
+    def: TypeDef,
```
holykol, line 110: Wait, is this intentional? Are we forbidding top level types to be anything but structs and enums? What about writing something like 
```
type ID = u32;
type Ids = Vec<ID>` 

``` 
for semantics?

idlparser/src/ffi/types/visitor.rs
DennisInSky: Tried, but unfortunatelly could not achieve it

PR #41, opened 2024-01-19
feat(wasm-ffi): Support FFI for WASM

DennisInSky:


idlparser/src/ffi/ast/visitor.rs
```diff
@@ -1,292 +1,479 @@
...
-    visit_struct_def: extern "C" fn(*mut Visitor, *const StructDef),
-    visit_struct_field: extern "C" fn(*mut Visitor, *const StructField),
-    visit_enum_def: extern "C" fn(*mut Visitor, *const EnumDef),
-    visit_enum_variant: extern "C" fn(*mut Visitor, *const EnumVariant),
-}
-
-/// # Safety
-///
-/// See documentation for [`const_ptr::as_ref`] and [`mut_ptr::as_mut`]
+    visit_service: unsafe extern "C" fn(context: *const (), *const Service),
```
holykol, line 8: Can we use `Ptr` struct in this file too? Never mind if I missed the point

holykol (approved):

idlparser/src/ffi/ast/visitor.rs
DennisInSky: We can, but it doesn't matter much as this Ptr is never touched by the parser

PR #53, opened 2024-01-25
Support wasm parser

osipov-mit:


js/src/parser/parser.wasm
DennisInSky, line 1: Why do we want to publish it onto git? Isn't it just a build artifact?

DennisInSky (approved):

js/src/parser/parser.wasm
osipov-mit: no need

Makefile
```diff
@@ -0,0 +1,9 @@
+build:
```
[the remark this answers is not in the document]
osipov-mit: renamed

DennisInSky (commented):
1. Generated client should be able to pass some value with every single call to a program.
2. There should be 2 promises whilest calling a program:
    - first one for sending a call message to the system, aka publishing the message
    - second one for receving the actual program response after it process the message

DennisInSky (approved):

PR #82, opened 2024-02-12
fix(sender): Correctly parse responses

holykol:


DennisInSky (changes requested):
If you don't check the response contains the exact name, it will be cheaper to store just the name length. It will require less mem allocations and computations

DennisInSky:
> If you don't check the response contains the exact name, it will be cheaper to store just the name length. It will require less mem allocations and computations

Agreed to keep the validation for the time being

DennisInSky (approved):

PR #119, opened 2024-03-05
feat(macros): Generate program metadata

### LINKED ISSUE #111: macros: Generate program meta
### Problem to Solve

The `gprogram` macro needs to generate meta data which can be used for generating program IDL

### Possible Solution

Introduce `ProgramMeta` trait which represents all required information making use of the scale-info's `MetaType`. Generate implementation using the `gprogram` macro for a specified program.

### Notes

_No response_

DennisInSky:
Resolves #111  .

This PR implements generating of program metadata which in turn will be used for generating program IDL. The metadata comprises of:
- metadata for program possible constructors
- metadata for services hosted by the program


macros/core/src/snapshots/sails_macros_core__program__tests__gprogram_generates_handle_for_multiple_services_with_empty_and_non_empty_routes.snap
```diff
@@ -11,6 +11,27 @@ impl MyProgram {
         MyService
     }
 }
+impl sails_idl_meta::ProgramMeta for MyProgram {
```
NikVolf, line 14: are those autogenerated? maybe exclude it from source contrrol?

DennisInSky: They are autogenerated at the very first run. If one is happy with the output, they accept them to be a reference. For the subsequent runs, if output differs, a test fails and a diff is shown.
https://crates.io/crates/insta

idl-meta/src/lib.rs
```diff
@@ -26,3 +26,30 @@ pub trait ServiceMeta {
     fn commands() -> MetaType;
     fn queries() -> MetaType;
 }
+
+pub struct AnyServiceMeta {
```
holykol, line 30: I think I miss the point of introducing new type. Is `ServiceMeta` not object safe or something?

holykol (approved):

idl-meta/src/lib.rs
DennisInSky: The point of this struct is to represent arbitrary service meta in program meta. The latter can have multiple services hosted by a program, so they a represented via an iterator. The `ServiceMeta` trait can't be used for iterator item, so this struct has got introduced

PR #121, opened 2024-03-11
feat(idl-gen): Generate constructor block

### LINKED ISSUE #113: idl-gen: Generate the `constructor` block
### Problem to Solve

Before using, program can optionally require to be constructed via some call. This has to be reflected in its IDL via an optional `constructor` block which consists of at least one possible methods, `constructor`

### Possible Solution

Use meta data generated by the `gprogram` macro in the scope of #111 and handlebars template to render it.

### Notes

_No response_

DennisInSky:
Resolves #113  .

Introduces code generating the `constructor` block in program IDL.
Also there is some refactoring with regards to moving meta data expansion from HBS code into rust one with some extra checks on meta invariants.


idlgen/src/type_names.rs
DennisInSky, line 1: It is just about renaming and replacing `PortableRegistry` by `impl Iterator<Item = &'a PortableType>`

holykol (approved):

PR #127, opened 2024-03-12
fix(macros): Remove unnecessary memory allocations during service invocation routing

### LINKED ISSUE #72: macros: Consider generating more optimal `requests::process`
### File Location(s)

currently it creates a lot of vectors just to find method
https://github.com/gear-tech/sails/blob/6d2d9cf38d4a5606b4d7963e8dd9af7142d9dbb6/macros/core/src/processors.rs#L62-L68

### Proposal

use some cheap hash function like [adler32](https://en.wikipedia.org/wiki/Adler-32) and create code like this:
```rust
match method_hash {
    0x044301b9 /*mint*/ => { /*pass rest payload as method args*/ },
    _ => panic!("unknown method"),
}
```

DennisInSky:
Resolves #72  .

Unnecessary memory allocations and runtime route encoding has been removed by this PR.
As it comes to using a hash in lieu of bytes, we might return to this later when we have some means allowing any node event listener like indexer to obtain an IDL for a program of interest and compute hash on its side so it can match responses with a specific methods.

### DIFF
--- a/macros/core/src/service.rs
+++ b/macros/core/src/service.rs
@@ -20,6 +20,7 @@
 
 use crate::shared::{self, Func, ImplType};
 use convert_case::{Case, Casing};
+use parity_scale_codec::Encode;
 use proc_macro2::{Span, TokenStream as TokenStream2};
 use proc_macro_error::abort;
 use quote::quote;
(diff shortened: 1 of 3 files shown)


macros/core/src/service.rs
```diff
@@ -63,13 +64,17 @@ pub fn gservice(service_impl_tokens: TokenStream2) -> TokenStream2 {
...
-            }
-        ));
+        invocation_dispatches.push({
+            let invocation_route_bytes = invocation_route.encode();
+            let invocation_route_len = invocation_route_bytes.len();
+            quote!(
+                if #input_ident.starts_with(& [ #(#invocation_route_bytes),* ]) {
+                    let output = self.#invocation_func_ident(&#input_ident[#invocation_route_len..]).await;
+                    static INVOCATION_ROUTE: [u8; #invocation_route_len] = [ #(#invocation_route_bytes),* ];
+                    return [INVOCATION_ROUTE.as_ref(), &output].concat();
```
StackOverflowExcept1on, line 74: 1. iirc `.concat()` also allocates memory
2. mb re-use `INVOCATION_ROUTE` in `.starts_with`?

StackOverflowExcept1on (approved):

PR #141, opened 2024-04-03
feat(rtl): Intorduce abstractions and some implementations for sending remote calls and program activations

DennisInSky:
Introduce abstractions and some implementation for executing remote calls in environment agnostic way.
The abstractions are supposed to be used by code generated from IDL (client) for interacting with on-chain programs.
The aim is to use the same client from different environments:
- gstd (program to program)
- gtest (gtest to program)
- gsdk (off-chain tool to program)
- mock (program interacting with mocked program)


rtl/src/calls.rs
```diff
@@ -0,0 +1,131 @@
...
+
+    fn with_args(self, args: TArgs) -> Self;
+
+    fn value(&self) -> ValueUnit;
+
+    fn args(&self) -> &TArgs;
+}
+
+pub struct CallTicket<TReplyFuture, TReply, const PREFIX_SIZE: usize> {
+    reply_prefix: &'static [u8; PREFIX_SIZE],
```
DennisInSky, line 22: It is fine to use just `&'static [u8]`

DennisInSky: done

holykol (approved):

PR #144, opened 2024-04-08
feat: Add support for new sails-rtl calls

holykol:
* Switch to sails-rtl for Call and Sender traits
* Remove sails-sender


client-gen/src/snapshots/sails_clientgen__generator__tests__basic_works.snap
```diff
@@ -6,31 +6,37 @@ expression: generate(program).unwrap()
 use parity_scale_codec::{Decode, Encode};
 #[allow(unused_imports)]
 use sails_rtl::collections::BTreeMap;
-use sails_rtl::{String, *};
-#[allow(unused_imports)]
-use sails_sender::{Call, CreateProgramCall, GStdSender};
+use sails_rtl::{
+    calls::RemotingAction, gstd::calls::Args, gstd::calls::Remoting, prelude::*, String,
```
DennisInSky, line 10: Generated code should remain generic against `Args` as it can be used for either environment, i.e. the code should not contain any references from the `sails_rtl::gstd` module

client-gen/src/snapshots/sails_clientgen__generator__tests__basic_works.snap
```diff
@@ -6,31 +6,37 @@ expression: generate(program).unwrap()
...
+};
 pub trait Service {
```
DennisInSky, line 12: This has to be a generic trait, exactly like in the snippet I sent over to you the other day

DennisInSky, line 12: Names for trait and its impl should reflect the name of the used IDL otherwise I will always need to use type aliasing. We discussed it earlier.

DennisInSky (changes requested):
You can raise separate issues for the comments, add links to them and them we merge this PR as is. There are 2 issues required:
1. Generate environment agnostic client
2. Beatify names for generated client trait and impl

client-gen/src/snapshots/sails_clientgen__generator__tests__basic_works.snap
holykol: Makes sense, implemented generating service name from idl file name

holykol: Take a look now

client-gen/src/snapshots/sails_clientgen__generator__tests__basic_works.snap
holykol: Take a look now

client-gen/src/snapshots/sails_clientgen__generator__tests__basic_works.snap
```diff
@@ -1,36 +1,43 @@
...
+    #[allow(clippy::type_complexity)]
+    fn do_this(&mut self, p1: u32, p2: MyParam) -> RemotingAction<Remoting, TCallArgs, u16>;
+    #[allow(clippy::type_complexity)]
+    fn do_that(&mut self, p1: (u8, u32)) -> RemotingAction<Remoting, TCallArgs, u8>;
 }
 
 #[derive(Clone)]
-pub struct Client {
-    sender: GStdSender,
+pub struct ServiceClient {
```
DennisInSky, line 21: Let's probably keep the same name as for the trait but move all this code into some `impl` sub-module. Another option is adding some `Impl` suffix like ServiceImpl. Don't know which way is more idiomatic for rustecians.

DennisInSky (changes requested):

DennisInSky (approved):

PR #157, opened 2024-04-18
feat: Move common build code to sails-builder

### LINKED ISSUE #95: Pack building commands into simple crate
instead of:
[this](https://github.com/gear-tech/sails/blob/268e11ac470a4f722eac79df0090b89e04fa4824/examples/this-that-svc/wasm/build.rs#L4-L15)

should be this:

```
sails_builder::new().generate_idl().build()
```

holykol:
Resolves #95

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -14,9 +14,9 @@ dependencies = [
 
 [[package]]
 name = "actor-system-error"
-version = "1.3.1"
+version = "1.4.0"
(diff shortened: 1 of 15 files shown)


DennisInSky (commented):
I would suggest moving this into draft for the time being as the final structure of crates and build flow is unclear yet

DennisInSky:
Outdated

PR #172, opened 2024-04-22
feat: Change generated clients to impl Call

holykol:
Keep in mind that generated traits are no longer object safe after this change


client-gen/src/snapshots/sails_clientgen__generator__tests__rmrk_works.snap
```diff
@@ -9,15 +9,15 @@ use parity_scale_codec::{Decode, Encode};
...
 #[allow(unused_imports)]
 use sails_rtl::{
-    calls::{Remoting, RemotingAction},
+    calls::{Activation, Call, Remoting, RemotingAction},
     prelude::*,
     String,
 };
 pub mod ctor {
     use super::*;
     pub trait RmrkCatalogConstructors<A: Default> {
```
DennisInSky, line 18: I think it has to be under the `traits` module with a name like `RmrkCatalogFactory`, i.e. it should be pretty much the same as for service client

examples/puppeteer/app/src/lib.rs
```diff
@@ -11,13 +11,16 @@ use sails_rtl::calls::Call;
...
+pub struct Puppeteer<ThisThatClient> {
+    puppet: ThisThatClient,
 }
 
 #[gservice]
-impl Puppeteer {
-    pub const fn new(puppet: Box<dyn ThisThatSvc<GStdRemoting, GStdArgs>>) -> Self {
+impl<ThisThatClient> Puppeteer<ThisThatClient>
+where
+    ThisThatClient: ThisThatSvc<GStdRemoting, GStdArgs>,
```
DennisInSky, line 21: Nit: Using generic doesn't seem make sense here because you require uinsg GStdRemoting, i.e. I won't be able to use this service with some mock of ThisThatSvc which doesn't use remoting at all. Said that, it doesn't make much sense to abtract it

client-gen/src/snapshots/sails_clientgen__generator__tests__rmrk_works.snap
```diff
@@ -9,72 +9,66 @@ use parity_scale_codec::{Decode, Encode};
...
-    }
-    pub struct ProgramFactory;
-    impl ProgramFactory {
-        #[allow(unused)]
-        pub fn new() -> Self {
-            Self
-        }
+pub struct RmrkCatalogFactory;
+impl RmrkCatalogFactory {
+    #[allow(unused)]
```
DennisInSky, line 18: Nit: Weird it is needed as it is a pub fn for pub struct

DennisInSky (approved):

client-gen/src/snapshots/sails_clientgen__generator__tests__rmrk_works.snap
holykol: checked again, removing `allow` still warns about unused function ¯\_(ツ)_/¯

PR #189, opened 2024-04-26
fix(client-gen): Pass paths with AsRef<Path>

### LINKED ISSUE #180: Use `AsRef<Path>` instead of `&Path` in API
### File Location(s)

`sails_clientgen::generate_client_from_idl`

### Proposal

Any API works with system paths uses `AsRef<Path>` in function signatures (check `std` for reference)

holykol:
Resolves #180

### DIFF
--- a/client-gen/src/lib.rs
+++ b/client-gen/src/lib.rs
@@ -4,7 +4,10 @@ use anyhow::{Context, Result};
 use convert_case::{Case, Casing};
 use std::{ffi::OsStr, fs, path::Path};
 
-pub fn generate_client_from_idl(idl_path: impl AsRef<Path>, out_path: impl AsRef<Path>) -> Result<()> {
+pub fn generate_client_from_idl(
+    idl_path: impl AsRef<Path>,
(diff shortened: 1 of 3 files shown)


client-gen/src/lib.rs
```diff
@@ -4,7 +4,10 @@ use anyhow::{Context, Result};
 use convert_case::{Case, Casing};
 use std::{ffi::OsStr, fs, path::Path};
 
-pub fn generate_client_from_idl(idl_path: &Path, out_path: &Path) -> Result<()> {
+pub fn generate_client_from_idl<P: AsRef<Path>>(idl_path: P, out_path: P) -> Result<()> {
```
DennisInSky, line 7: This code requires P to be the same type for both arguments which might not be the case, so it is better to write
```suggestion
pub fn generate_client_from_idl(idl_path: impl AsRef<Path>, out_path: impl AsRef<Path>) -> Result<()> {
```

DennisInSky (approved):

PR #196, opened 2024-04-29
fix(events): Prefix emitted events with service invocation route

DennisInSky:
Each emitted event payload needs be prefixed with service name/invocation route. Event trigger resolves the latter via a map which is maintained by dispatching code which attributes id of being processed message with the route. The approach requires at least 2 calls to `gstd::msg::id()` which is not that cheap. Current there is no option to cache it somehow.


DennisInSky:
@osipov-mit FYI. It might affect your code where you parse event payload

PR #201, opened 2024-05-01
feat(idlgen): Support omitting return type

### LINKED ISSUE #55: idl-gen: Support service methods with omitted return type
### Problem to Solve

Service methods w/o explicit return types
`do_something(&mut self) {}`
are not supported atm

### Possible Solution

Service methods w/o explicit return types
`do_something(&mut self) {}`
should be treated as returning `()`

### Notes

_No response_

holykol:
Resolves #55

### DIFF
--- a/examples/this-that-svc/wasm/this-that-svc.idl
+++ b/examples/this-that-svc/wasm/this-that-svc.idl
@@ -20,6 +20,7 @@ type TupleStruct = struct {
 service {
   DoThat : (param: DoThatParam) -> result (struct { str, u32 }, struct { str });
   DoThis : (p1: u32, p2: str, p3: struct { opt str, u8 }, p4: TupleStruct) -> struct { str, u32 };
+  Noop : () -> null;
   query That : () -> result (str, str);
   query This : () -> u32;
 };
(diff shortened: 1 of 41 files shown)


examples/this-that-svc/wasm/this-that-svc.idl
```diff
@@ -20,6 +20,7 @@ type TupleStruct = struct {
 service {
   DoThat : (param: DoThatParam) -> result (struct { str, u32 }, struct { str });
   DoThis : (p1: u32, p2: str, p3: struct { opt str, u8 }, p4: TupleStruct) -> struct { str, u32 };
+  Noop : ();
```
DennisInSky, line 23: Not quite right. The behaviour should be similar to processing a func of the following shape:
```
fn omit_result_type() -> () {}
```
Which will result in:
```
OmitResultType: () -> null;
```
In order to address the issue all is needed is adjusting this code here
https://github.com/gear-tech/sails/blob/a73a73dad9d1fc847a834400500acee04d2e75f8/macros/core/src/shared.rs#L116
Matching against `ReturnType::Default` and returning `()` as Type
https://docs.rs/syn/latest/syn/enum.ReturnType.html#variant.Default

DennisInSky (changes requested):

examples/this-that-svc/wasm/this-that-svc.idl
DennisInSky: All the other code will remain untouched after that.

DennisInSky (approved):

PR #205, opened 2024-05-03
fix: Resolve lints for Rust 1.78

holykol:


rtl/src/prelude.rs
```diff
@@ -41,7 +41,8 @@ pub mod collections {
 pub mod ffi {
     extern crate alloc;
 
-    pub use alloc::ffi::*;
+    pub use alloc::ffi::CString;
+    pub use alloc::ffi::{FromVecWithNulError, IntoStringError, NulError};
```
DennisInSky, line 45: Nit
```suggestion
    pub use alloc::ffi::{CString, FromVecWithNulError, IntoStringError, NulError};
```

DennisInSky (approved):

rtl/src/prelude.rs
holykol: thx

PR #251, opened 2024-05-15
refactor(rtl): Make sails-rtl crate self-sufficient

### LINKED ISSUE #182: Re-export `scale-info` and `parity-scale-info` in procedural macros
### File Location(s)

_No response_

### Proposal

`#[gprogram]` and `#[gservice]` require user has these dependencies in his `Cargo.toml`. They can be just re-exported

DennisInSky:
With this PR a user need to wire up just the `sails-rtl` crate in order to build an empty sails-based program:
1. The `gprogram`, `gservice` and `groute` macros are re-exported
2. The generated code relies on re-exported `parity-scale-codec` and `scale-info` crates

Resolves #182  .

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3402,7 +3402,6 @@ dependencies = [
  "gstd",
  "parity-scale-codec",
  "sails-client-gen",
- "sails-macros",
  "sails-rtl",
  "scale-info",
 ]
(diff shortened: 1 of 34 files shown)


holykol (approved):
nice

PR #256, opened 2024-05-17
feat(examples): Add example returning references

### LINKED ISSUE #206: macros: Allow handler return types to be a reference
### Problem to Solve

At the moment the return type must be an owned type only, this is not efficient from perf point of view as the only thing happening to the result is its encoding which can be successfully done via references

### Possible Solution

Support ref types in as result of service handler. The behaviour should be similar to the owned types, i.e. the representation in IDL and in client generated code

### Notes

_No response_

holykol:
related to #206


DennisInSky:
@breathx what kind of references would you expect to be supported? Taking account of stateless nature of services, would it be enough only references with static lifetimes are supported?

breathx:
> @breathx what kind of references would you expect to be supported? Taking account of stateless nature of services, would it be enough only references with static lifetimes are supported?

Discussed. Agreed on static references atm

DennisInSky (approved):

PR #257, opened 2024-05-17
fix(macros): unexpected input handler produces wrong output

### LINKED ISSUE #238: macros: Unexpected input handler produces wrong output
### Problem

If unexpected input can't be decoded as a string, the panic message output displays empty payload
https://github.com/gear-tech/sails/blob/3e799074d619e9dcd20a0f9ce4637d60d999a731/macros/core/src/shared.rs#L159

### Steps

1. Send unexpected payload to a program, i.e. one which doesn't target any registered service

### Possible Solution

_No response_

### Notes

_No response_

### Relevant Log Output

<details><summary>Click to expand/collapse</summary>
<p>

```
<Paste your log here>
```

</p>
</details>

vobradovich:
Resolves #238  .

Copy input pointer before `String::decode`
```rust
        let mut #copy_ident = #input_ident;
        let input = String::decode(&mut #copy_ident)
```


Cargo.toml
```diff
@@ -29,6 +29,7 @@ members = [
 anyhow = "1"
 const-format = { package = "const_format", version = "0.2" }
 convert-case = { package = "convert_case", version = "0.6" }
+fluent-asserter = "0.1.9"
```
DennisInSky, line 32: Require "0.1" should be enough
```suggestion
fluent-asserter = "0.1"

```

DennisInSky (approved):

DennisInSky (approved):

PR #258, opened 2024-05-20
feat(js): support vouchers in transaction builder

osipov-mit:


examples/ping/app/src/lib.rs
```diff
@@ -0,0 +1,21 @@
...
+#[derive(Default)]
+pub struct Program;
+
+#[gprogram]
+impl Program {
+    pub fn new() -> Self {
+        Self
+    }
+
+    #[groute("Ping")]
```
DennisInSky, line 17: Nit: can be removed

examples/ping/app/src/service.rs
```diff
@@ -0,0 +1,24 @@
...
+            Ok("pong".into())
+        }
+    }
+}
+
+impl Default for PingService {
+    fn default() -> Self {
+        Self::new()
+    }
+}
```
DennisInSky, line 24: #[derive(Default)] should be enough

DennisInSky (approved):
Checked the program in examples only

Issue #183, opened 2024-04-23
rustc complains about unused fields in `__NewParams`


ark0f:
### Problem

rustc complains about unused fields in `__NewParams` because fields are used only in WASM but they aren't in host target

### Steps

```rust
#[gprogram]
impl Program {
    pub fn new(config: AnyThing) -> Self {
		let _config = config;
        Self
    }
}
```

vobradovich:
Exact output
```
warning: field `config` is never read
 --> C:\git\sails\macros\tests\ui\gprogram_works.rs:7:16
  |
5 | #[gprogram]
  | ----------- field in this struct
6 | impl MyProgram {
7 |     pub fn new(config: u32) -> Self {
  |                ^^^^^^
  |
  = note: `#[warn(dead_code)]` on by default
  ```

PR #259, opened 2024-05-20
fix(macros): rustc complains about unused fields in __NewParams, allow dead_code on generated struct

vobradovich:
Resolves https://github.com/gear-tech/sails/issues/183 .

### DIFF
--- a/macros/tests/ui/gprogram_works.rs
+++ b/macros/tests/ui/gprogram_works.rs
@@ -36,5 +36,4 @@ pub mod svc {
 #[tokio::main]
 async fn main() {
     let _prg = MyProgram::new(42);
-    // let _s  = &prg.service1(); // panic here (!)
 }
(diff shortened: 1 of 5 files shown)


macros/tests/ui/gprogram_works.rs
```diff
@@ -0,0 +1,39 @@
...
+        Self
+    }
+
+    pub fn service1(&self) -> svc::MyService {
+        svc::MyService
+    }
+}
+
+pub mod svc {
+    #![allow(dead_code)]
```
DennisInSky, line 18: Why do we need this one?

DennisInSky (approved):

macros/tests/ui/gprogram_works.rs
vobradovich: To silent the compiler, service methods are not used in test. Check only gprogram.

PR #281, opened 2024-05-23
feat: Service enxtending

DennisInSky:
This PR introduces a new feature which allows a service to extend another service via mentioning it in the `gservice` attribute of the former: `#[gservice(extends = BaseService1)]` or `#[gservice(extends = [BaseService1, BaseService2]]`.

The extending means that all methods and events of the base services will be exposed as methods and events of the target service, i.e. they will be visible in the IDL of the target service as they would belong to it, the messages will be dispatched accrodingly.

In case of method names overlapping, the first observed method is taken. The order of observation is as follows: target service then base services in order of their specification in the `gservice` attribute. This behaviour alows method overriding.

In case of event names overlapping, an error about ambiguous names will be reported by IDL generator. Unfortunatelly, it is not possible to detect this situation at compile time.


macros/core/src/service.rs
```diff
@@ -1,413 +1,554 @@
...
+        (
+            service_type.path().clone(),
+            service_type.args().clone(),
+            service_type.constraints().cloned(),
+        )
+    };
+
+    let service_handlers = discover_service_handlers(&service_impl);
+
+    if service_handlers.is_empty() {
```
vobradovich, line 113: Can service only expose other services without having  its own impl?

DennisInSky: Good question. I think I'll make it so a service should have its own methods or extend at least one service, or both

DennisInSky: Adjusted

vobradovich (approved):

Issue #200, opened 2024-05-01
tests: Make use of generated RmrkResource client in integration tests


DennisInSky:
### Proposal

At the moment integration tests use raw functionality to interact with programs in `GTest` environment, i.e. they compose payload themselves. It should be possible to make use of generated client code in addition with `GTestRemoting` (`GTestProgramSpace`)

PR #282, opened 2024-05-23
tests: Make use of generated RmrkResource client in integration tests

vobradovich:
Resolves https://github.com/gear-tech/sails/issues/200

### DIFF
--- a/examples/rmrk/resource/wasm/Cargo.toml
+++ b/examples/rmrk/resource/wasm/Cargo.toml
@@ -18,4 +18,5 @@ parity-scale-codec = { workspace = true, features = ["derive"] }
 rmrk-catalog = { path = "../../catalog/wasm" }
 sails-rtl.workspace = true
 scale-info = { workspace = true, features = ["derive"] }
+thiserror-no-std.workspace = true
 tokio = { workspace = true, features = ["rt", "macros"] }
--- a/examples/rmrk/resource/wasm/Cargo.toml
+++ b/examples/rmrk/resource/wasm/Cargo.toml
@@ -9,10 +9,13 @@ rmrk-resource-app = { path = "../app" }
 [build-dependencies]
 gwasm-builder.workspace = true
 rmrk-resource-app = { path = "../app" }
+sails-client-gen.workspace = true
 sails-idl-gen.workspace = true
 
 [dev-dependencies]
 gtest.workspace = true
+parity-scale-codec = { workspace = true, features = ["derive"] }
 rmrk-catalog = { path = "../../catalog/wasm" }
 sails-rtl.workspace = true
+scale-info = { workspace = true, features = ["derive"] }
 tokio = { workspace = true, features = ["rt", "macros"] }
--- a/examples/rmrk/resource/wasm/tests/resources.rs
+++ b/examples/rmrk/resource/wasm/tests/resources.rs
@@ -1,3 +1,4 @@
+use crate::rmrk_resource_client::traits::RmrkResource;
 use core::cell::OnceCell;
 use gtest::{Program, RunResult};
 use rmrk_catalog::services::parts::{FixedPart, Part};
@@ -7,12 +8,14 @@ use rmrk_resource_app::services::{
     ResourceStorageEvent,
 };
 use sails_rtl::{
-    calls::Remoting,
+    calls::{Call, Remoting},
     collections::BTreeMap,
     errors::Result,
     gtest::calls::{GTestArgs, GTestRemoting},
     ActorId, Decode, Encode,
 };
+
+mod rmrk_resource_client;
 
 const CATALOG_PROGRAM_WASM_PATH: &str =
     "../../../../target/wasm32-unknown-unknown/debug/rmrk_catalog.wasm";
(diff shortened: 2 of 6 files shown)


client-gen/src/generator.rs
```diff
@@ -198,6 +198,7 @@ impl<'ast> Visitor<'ast> for CtorFactoryGenerator {
         self.code.push_str(
             "
             #[allow(unused)]
+            #[allow(clippy::new_without_default)]
```
DennisInSky, line 201: Let's add:
1. `#[derive(Default)]` to the factotry itself
2. Make it `pub struct {}Factory(())` so it can't be constructed from the outside unless when it is done either via default or new

examples/rmrk/resource/wasm/tests/rmrk_resource_client.rs
DennisInSky, line 1: I would suggest making this as a sub-mod of the `resources` as for integration tests every root module is treated as a separate set of tests

examples/rmrk/resource/wasm/Cargo.toml
```diff
@@ -9,10 +9,14 @@ rmrk-resource-app = { path = "../app" }
 [build-dependencies]
 gwasm-builder.workspace = true
 rmrk-resource-app = { path = "../app" }
+sails-client-gen.workspace = true
 sails-idl-gen.workspace = true
 
 [dev-dependencies]
 gtest.workspace = true
+parity-scale-codec = { workspace = true, features = ["derive"] }
```
DennisInSky, line 17: This shouldn't be needed. If it is, then there is some issue which needs be fixed. The same is about `scale-info`. I can guess the issue is in the `client-gen` which doesn't make use of `#[codec(crate = sails_rtl::scale_codec)]` and `#[scale_info(crate = sails_rtl::scale_info)]` for re-exported stuff

examples/rmrk/resource/wasm/tests/resources.rs
```diff
@@ -195,6 +199,87 @@ fn adding_non_existing_part_to_resource_fails() {
     assert!(run_result.contains(&(ADMIN_ID, expected_response)));
 }
 
+#[tokio::test]
+async fn adding_resource_to_storage_using_client_succeeds() {
```
DennisInSky, line 203: Can we please renaming it so it matches the other similar methods `adding_resource_to_storage_by_admin_via_client_succeeds` and place it next to those methods so it's seen how the same functionality is tested using different means

examples/rmrk/resource/wasm/tests/resources.rs
```diff
@@ -373,4 +464,55 @@ impl<'a> Fixture<'a> {
         .concat();
         program.send_bytes(actor_id, encoded_request)
     }
+
+    async fn add_resource_using_client(
```
DennisInSky, line 468: Let's move it next to the similar method please

examples/rmrk/resource/wasm/tests/resources.rs
```diff
@@ -373,4 +464,55 @@ impl<'a> Fixture<'a> {
...
+        &'a self,
+        actor_id: ActorId,
```
DennisInSky, line 470: Let's keep it `u64` in order to avoid ugly code in the test itself

examples/rmrk/resource/wasm/tests/resources.rs
```diff
@@ -373,4 +464,55 @@ impl<'a> Fixture<'a> {
...
+        resource_id: u8,
+        resource: rmrk_resource_client::Resource,
+    ) -> Result<(u8, rmrk_resource_client::Resource), TestError> {
+        let mut resource_client = self.resource_client();
+        let call = resource_client
+            .add_resource_entry(resource_id, resource)
+            .with_args(GTestArgs::new(self.admin_id.into()))
+            .publish(actor_id)
```
DennisInSky, line 478: Let's rename the `actor_id` param everywhere as it confuses. I guess this what tripped you up in the first place

DennisInSky (changes requested):

examples/rmrk/resource/wasm/tests/resources.rs
vobradovich: renamed to `program_id: ActorId`

client-gen/src/generator.rs
```diff
@@ -191,15 +190,16 @@ impl CtorFactoryGenerator {
 
 impl<'ast> Visitor<'ast> for CtorFactoryGenerator {
     fn visit_ctor(&mut self, ctor: &'ast Ctor) {
+        self.code.push_str("#[derive(Default)]");
         self.code
-            .push_str(&format!("pub struct {}Factory;\n", self.service_name));
+            .push_str(&format!("pub struct {}Factory(());\n", self.service_name));
```
holykol, line 195: Is this required? Isn't `Factory` the same thing as `Factory(())`?

holykol (approved):
nice

examples/rmrk/resource/wasm/tests/rmrk_resource_client/mod.rs
DennisInSky, line 1: nit: Adding yet another `rmrk` suffix seems excessive. `resource_client` should be enough

examples/rmrk/resource/wasm/tests/resources.rs
```diff
@@ -115,6 +118,31 @@ async fn adding_resource_to_storage_by_admin_succeeds_async() {
...
+    // Act
+    let resource = resource_client::Resource::Composed(resource_client::ComposedResource {
+        src: "<src_uri>".into(),
+        thumb: "<thumb_uri>".into(),
+        metadata_uri: "<metadata_uri>".into(),
+        base: catalog_program_id,
+        parts: vec![1, 2, 3],
+    });
+    let add_reply = fixture
+        .add_resource_via_client(ADMIN_ID, RESOURCE_ID, resource, resource_program_id)
```
DennisInSky, line 137: nit: resource_program_id can be taken and passed inside the method as you retrieve it from the fixture anyway

examples/rmrk/resource/wasm/tests/resources.rs
```diff
@@ -115,6 +118,31 @@ async fn adding_resource_to_storage_by_admin_succeeds_async() {
...
+        .await
+        .unwrap()
+        .unwrap();
+
+    // Assert
+    assert_eq!(RESOURCE_ID, add_reply.0);
```
DennisInSky, line 143: nit: I would suggest asserting `assert_eq(Ok(RESOURCE_ID), apply_reply.map(|r| r.0))` instead of unwrapping at line 140

examples/rmrk/resource/wasm/tests/resources.rs
```diff
@@ -165,6 +193,63 @@ fn adding_existing_part_to_resource_by_admin_succeeds() {
...
+    let add_part_reply = fixture
+        .add_part_to_resource_via_client(ADMIN_ID, RESOURCE_ID, PART_ID, resource_program_id)
+        .await
+        .unwrap()
+        .unwrap();
+
+    // Assert
+    assert_eq!(PART_ID, add_part_reply);
+
+    // Act
```
DennisInSky, line 235: This is a part of the assert as test name suggests

examples/rmrk/resource/wasm/tests/resources.rs
```diff
@@ -165,6 +194,62 @@ fn adding_existing_part_to_resource_by_admin_succeeds() {
     }
 }
 
+#[tokio::test]
+async fn adding_existing_part_to_resource_by_admin_via_client_succeeds() {
+    // Arrange
+    let fixture = Fixture::new(ADMIN_ID);
+    let catalog_program_id = ActorId::from(fixture.catalog_program().id().into_bytes());
+    let resource_program_id = ActorId::from(fixture.resource_program_for_async().id().into_bytes());
```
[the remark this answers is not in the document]
DennisInSky: the same comment about using of the `resource_program_id` is applicable here

examples/rmrk/resource/wasm/tests/resources.rs
```diff
@@ -165,6 +193,63 @@ fn adding_existing_part_to_resource_by_admin_succeeds() {
...
+
+    // Act
+    let add_part_reply = fixture
+        .add_part_to_resource_via_client(ADMIN_ID, RESOURCE_ID, PART_ID, resource_program_id)
+        .await
+        .unwrap()
+        .unwrap();
+
+    // Assert
+    assert_eq!(PART_ID, add_part_reply);
```
DennisInSky, line 233: Instead of unwrapping at line 230
```suggestion
    assert_eq!(Ok(PART_ID), Ok(add_part_reply));
```

examples/rmrk/resource/wasm/tests/resources.rs
```diff
@@ -1,3 +1,4 @@
+use crate::resource_client::traits::RmrkResource;
```
DennisInSky, line 1:
```suggestion
use crate::resource_client::traits::RmrkResource as RmrkResourceClient;
```

examples/rmrk/resource/wasm/tests/resources.rs
```diff
@@ -219,6 +304,10 @@ impl<'a> Fixture<'a> {
         &self.program_space
     }
 
+    fn resource_client(&self) -> crate::resource_client::RmrkResource<GTestRemoting, GTestArgs> {
```
DennisInSky, line 307: you can even add a type alias for RmrkResourceClient in order to avoid writting generic params all the time
```suggestion
    fn resource_client(&self) -> RmrkResourceClient<GTestRemoting, GTestArgs> {
```

vobradovich: added type alias RmrkResourceClient

Issue #147, opened 2024-04-08
idlparser: FFI error handling and panic safety


ark0f:
### File Location(s)

`idlparser::ffi` module

### Proposal

Currently `parse_idl` just unwraps any error and doesn't handle possible Rust code panic which can cause UB

PR #283, opened 2024-05-23
feat(js): Add error handling during IDL parsing

holykol:
closes https://github.com/gear-tech/sails/issues/147


idl-parser/src/ffi/ast/mod.rs
```diff
@@ -1,18 +1,43 @@
...
+pub unsafe extern "C" fn parse_idl(idl_ptr: *const u8, idl_len: u32) -> *mut ParseResult {
+    let idl_slice = unsafe { slice::from_raw_parts(idl_ptr, idl_len as usize) };
+    let idl_str = match str::from_utf8(idl_slice) {
+        Ok(s) => s,
+        Err(e) => return create_error(e, "validate IDL string"),
+    };
+
+    let program = match ast::parse_idl(idl_str) {
+        Ok(p) => p,
+        Err(e) => return create_error(e, "parse IDL"),
```
DennisInSky, line 29: Apparently, memory allocated for the error message will leak. Should we just return some static text?

DennisInSky (changes requested):
Did you want to cover just this case by this PR?
There are plenty of other `unwrap`s, etc. in exposed funcs

idl-parser/src/ffi/ast/mod.rs
```diff
@@ -1,18 +1,43 @@
...
+    slice, str,
+};
 
 pub mod visitor;
 
+#[repr(C)]
+pub enum ParseResult {
+    Success(*mut Program),
+    Error(*const c_char),
+}
```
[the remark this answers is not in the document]
holykol: `#[repr(C)]` is important in FFI context. `std::result::Result` does not have this tag, so it's memory layout is not reliable

see https://doc.rust-lang.org/nomicon/repr-rust.html and https://doc.rust-lang.org/nomicon/other-reprs.html#reprc

holykol:
> Did you want to cover just this case by this PR? There are plenty of other `unwrap`s, etc. in exposed funcs

I think seeing parsing error during idl generation is pretty important. We can deal with other unwraps in separate PR's, although I don't see what we can do about them besides just returning None/error when we receive null pointers

idl-parser/src/ffi/ast/mod.rs
holykol: Dynamic string will provide more information about error, including line:col of where parsing error happened. I've changed free logic to deallocate result instead of just program

DennisInSky: Hmm... But rust doesn't provide safe ffi for enums with fields. You can achieve it via using an intermediate enum for tagging, but it doesn't seem nice. Would it be more conventional to introduce yet another method for getting last error which takes a user passed buffer for error details as freeing error message via a separate call doesn't look common practice either

DennisInSky:
> > Did you want to cover just this case by this PR? There are plenty of other `unwrap`s, etc. in exposed funcs
> 
> I think seeing parsing error during idl generation is pretty important. We can deal with other unwraps in separate PR's, although I don't see what we can do about them besides just returning None/error when we receive null pointers

Of course seeing this information is important, but the main goal of the issue is providing well behaving API w/o UBs.

DennisInSky (commented):
My proposal is:
1. We introduce an FFI struct and enum for errors comprising of 2 fields:
```
enum ErrorCode {
    Ok = 0,
    ...
}

struct Error {
    code: ErrorCode,
    details: *const c_char,
}
```
2. For the `parse_idl` we introduce another struct
```
struct ParseResult {
    program: *mut Program,
    error: Error,
}
```
3. For all exposed `accept` funcs we return the `Error` struct

DennisInSky:
> My proposal is:
> 
> 1. We introduce an FFI struct and enum for errors comprising of 2 fields:
> 
> ```
> enum ErrorCode {
>     Ok = 0,
>     ...
> }
> 
> struct Error {
>     code: ErrorCode,
>     details: *const c_char,
> }
> ```
> 
> 2. For the `parse_idl` we introduce another struct
> 
> ```
> struct ParseResult {
>     program: *mut Program,
>     error: Error,
> }
> ```

This struct needs be freed via the `free_parse_result` func

> 
> 3. For all exposed `accept` funcs we return the `Error` struct

Here we never return any details unless they are a static string

DennisInSky (commented):
Agrred to re-do using struct instead of enum as suggested in proposal

idl-parser/src/ffi/ast/mod.rs
```diff
@@ -1,30 +1,127 @@
...
+#[repr(C)]
+pub struct ParseResult {
+    program: *mut Program,
+    error: Error,
+}
+
+#[repr(C)]
+pub struct AcceptResult {
+    error: Error,
+}
```
DennisInSky, line 18: I would return just Error w/o introducing any result as there is no such

idl-parser/src/ffi/ast/mod.rs
```diff
@@ -1,30 +1,127 @@
...
+    let details = CString::new(format!("{}: {}", context, e)).unwrap();
+    let result = ParseResult {
+        error: Error {
+            code,
+            details: details.into_raw(),
+        },
+        program: std::ptr::null_mut(),
+    };
+    Box::into_raw(Box::new(result))
+}
```
DennisInSky, line 76: nit : I would inline it

idl-parser/src/ffi/ast/sizes.txt
DennisInSky, line 1: Why do we need this?

idl-parser/src/ffi/ast/mod.rs
```diff
@@ -1,30 +1,127 @@
...
 }
 
 /// # Safety
 ///
-/// TODO
+/// Pointer must be obtained from `accept_*` function
 #[no_mangle]
-pub unsafe extern "C" fn free_program(program: *mut Program) {
-    if program.is_null() {
+pub unsafe extern "C" fn free_accept_result(result: *mut AcceptResult) {
```
DennisInSky, line 118: Can't we just return Error as a struct w/o doing any allocations/deallocatons. I would even say we can return just ErrorCode as we never populate details

idl-parser/src/ffi/ast/mod.rs
```diff
@@ -1,30 +1,127 @@
...
+use std::{
+    ffi::{c_char, CString},
+    slice, str,
+};
 
 pub mod visitor;
 
+#[repr(C)]
+pub struct ParseResult {
+    program: *mut Program,
```
DennisInSky, line 11: Btw, why do we need a pointer here? Isn't allocating just `ParseResult` enough?

holykol: It points to Program, which is not repr(C). So having it in a separate pointer makes sense

idl-parser/src/ffi/ast/mod.rs
holykol: started to return ErrorCode directly

DennisInSky: Waiting for the changes then

DennisInSky (commented):
> started to return ErrorCode directly
Waiting for this to happen

holykol:
oops, forgot to push

idl-parser/src/ffi/ast/visitor.rs
```diff
@@ -85,81 +85,131 @@ static VISITOR: Visitor = Visitor {
     visit_enum_variant,
 };
 
+macro_rules! defer {
```
DennisInSky, line 88: Should it be `deref` instead of `defer`? Or even `deref_data_ptr` so it matches to `deref_visitor`

idl-parser/src/ffi/ast/visitor.rs
```diff
@@ -85,81 +85,131 @@ static VISITOR: Visitor = Visitor {
...
+    ($expr:expr) => {{
+        if $expr.is_null() {
+            return ErrorCode::NullPtr;
+        }
+        unsafe { $expr.as_ref() }.unwrap()
+    }};
+}
+
+macro_rules! create_visitor {
```
DennisInSky, line 97:
```suggestion
macro_rules! deref_visitor {
```

DennisInSky (commented):
Just a few renaming suggestions

idl-parser/src/ffi/ast/visitor.rs
holykol: thanks, that was a typo

DennisInSky (approved):
Cool

PR #294, opened 2024-05-29
feat(client-gen): Introduce scv_io module

holykol:
- remove old *Svc*CallBuilder struct
- introduce new *svc_name*_calls module
- each *MethodName*Call is an tuple of arguments of the call, that can be encoded and decoded using encode() and from_bytes() methods
- each Call also has `decode_reply` static method, which can be used to parse reply of given method

considerations:
- is there a better way to organize this? (see `test_call_module` for example usage)
- maybe make *MethodName*Call fields public and add constructor method?


DennisInSky (commented):
Why can't we have it like we discussed, i.e. we generate:
- a module `<service_name>_io`
- inside the module we generate a struct `DoThis(())`
- the struct has 2 associated methods:
```
impl DoThis {
    pub fn encode_call(p1, p2, p3) -> Vec<u8> { ... }
    pub fn decode_reply(bytes: &[u8]) -> Result<ReturnType> { ... }
}
```
where the `ReturnType` is the actual type the call returns

client-gen/tests/snapshots/generator__rmrk_works.snap
```diff
@@ -146,90 +146,190 @@ where
     }
 }
 
-#[derive(Default)]
-pub struct RmrkCatalogCallBuilder;
+pub mod rmrk_catalog_calls {
```
DennisInSky, line 149: I would suggest naming it just `io` - it is what they are used to now + generated client is already in a mod like `puppet` so `puppet::io` is quite intuitive

client-gen/tests/snapshots/generator__rmrk_works.snap
```diff
@@ -146,90 +146,190 @@ where
...
+                56, 65, 100, 100, 69, 113, 117, 105, 112, 112, 97, 98, 108, 101, 115,
+            ]);
+            args.encode_to(&mut result);
+            result
+        }
+        #[allow(unused)]
+        pub fn decode_reply(
+            mut reply: &[u8],
+        ) -> Result<Result<(u32, Vec<ActorId>), Error>, sails_rtl::errors::Error> {
+            let result = Decode::decode(&mut reply).map_err(sails_rtl::errors::Error::Codec)?;
```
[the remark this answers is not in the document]
holykol: added a test

client-gen/tests/snapshots/generator__rmrk_works.snap
```diff
@@ -146,90 +146,190 @@ where
...
-        result.extend_from_slice(&[
-            56, 65, 100, 100, 69, 113, 117, 105, 112, 112, 97, 98, 108, 101, 115,
-        ]);
-        args.encode_to(&mut result);
-        result
+    #[derive(Debug, Clone, Copy)]
+    pub struct AddEquippablesCall(());
+
+    impl AddEquippablesCall {
+        #[allow(unused)]
```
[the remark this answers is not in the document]
holykol: for some reason clippy still complains about this, so I've replaced `allow(unused)` with `allow(dead_code)`

client-gen/tests/snapshots/generator__rmrk_works.snap
holykol: renamed

client-gen/tests/snapshots/generator__basic_works.snap
```diff
@@ -49,27 +49,59 @@ where
     }
 }
 
-#[derive(Default)]
-pub struct BasicCallBuilder;
+pub mod basic_io {
+    use super::*;
+
+    #[derive(Debug, Default, Clone, Copy)]
+    pub struct DoThisCall(());
```
DennisInSky, line 56: Let's omit the `Call` suffix. What do you think?

DennisInSky (approved):

PR #307, opened 2024-06-04
doc: Sails basic idea and concepts

DennisInSky:


DennisInSky:
While I was writing this doc I got a feeling that we might want to rename the `gprogram` macro into `gapp` so instead of using the word `application` when referring to a program as a whole not as a concept I could use program. Otherwise it gets a bit confusing - program for your program :-)

holykol (approved):
Nice work. Imo we definitely need to write not just "getting started", but "getting started with JS client".

NikVolf (approved):

DennisInSky:
@holykol thanks for the comments. I have created this issue #316 and consider them later when we accumulate more feedback on the doc. Meanwhile I merge it as it is

Issue #306, opened 2024-06-04
idl-parser: IDL with struct field names matching keywords can't be parsed


DennisInSky:
### Problem

If a struct type contains field with names matching keywords like `result`, `vec`, `service`, etc., it fails to get parsed

### Steps

```
struct { step: u8, result: u8 };
```

### Possible Solution

Prefix all such fields with `@` during IDL generation, and remove all leading `@` during IDL parsing

### Relevant Log Output

<details><summary>Click to expand/collapse</summary>
<p>

```
<Paste your log here>
```

</p>
</details>

PR #308, opened 2024-06-07
fix(idl-parser): Allow reserved keywords in struct fields & func params

vobradovich:
Resolves # https://github.com/gear-tech/sails/issues/306.


DennisInSky (approved):

DennisInSky:
@osipov-mit FYI. All the keywords are now accepted as field names as well as func param names

Issue #309, opened 2024-06-07
feat(macros): Propagate params handle_reply, handle_signal from grpogram to async_main


vobradovich:
### Problem to Solve

gstd::async_main allow to specify custom `handle_reply` and `handle_signal` handlers.
```rust
#[gstd::async_main(handle_reply = my_handle_reply, handle_signal = my_handle_signal)]
```

### Possible Solution

Allow to specify custom `handle_reply` and `handle_signal` handlers in `gprogram`

PR #312, opened 2024-06-10
feat(macros): Propagate params handle_reply, handle_signal from grpogram to async_main

vobradovich:
Resolves # https://github.com/gear-tech/sails/issues/309.


macros/core/src/program/mod.rs
```diff
@@ -12,19 +13,33 @@ use syn::{
...
-    gen_gprogram_impl(program_impl)
+    let args = parse_args(args);
+    gen_gprogram_impl(program_impl, args)
+}
+
+fn parse_args(args: TokenStream2) -> ProgramArgs {
+    syn::parse2(args).unwrap_or_else(|err| {
+        abort!(
+            err.span(),
+            "`gprogram` attribute can only contain `handle_reply` and `handle_signal` parameters: {}",
```
DennisInSky, line 39: How do you know here what went wrong? To my POV, this error should be raise in place where it has appropriate context

macros/core/src/program/mod.rs
```diff
@@ -332,8 +349,21 @@ fn generate_handle<'a>(
...
 
+    let mut args = Vec::with_capacity(2);
+    if let Some(handle_reply) = program_args.handle_reply() {
+        args.push(quote!(handle_reply = #handle_reply));
+    }
+    if let Some(handle_signal) = program_args.handle_signal() {
+        args.push(quote!(handle_signal = #handle_signal));
+    }
+    let async_main_args = if args.is_empty() {
+        quote!()
```
DennisInSky, line 360: Doesn't make much sense, the below quote will result in nothing if `args` is empty

macros/core/src/program/args.rs
```diff
@@ -0,0 +1,115 @@
...
+            }
+
+            match &*name {
+                "handle_reply" => {
+                    attrs.handle_reply = Some(path);
+                }
+                "handle_signal" => {
+                    attrs.handle_signal = Some(path);
+                }
+                _ => return Err(syn::Error::new_spanned(name, "unknown parameter")),
```
DennisInSky, line 46: I would abort here as this is the place where we know the proper error context

macros/core/src/program/mod.rs
vobradovich: it is for generating `#[gstd::async_main]` without `()`.

DennisInSky: Does it matter? Can't it have empty braces?

macros/core/src/program/args.rs
vobradovich: Spanned error is returned here

macros/core/src/program/mod.rs
vobradovich: Takes span from the error

macros/core/src/program/mod.rs
vobradovich: you can) just to minimize changes in insta files with this PR

DennisInSky (approved):

PR #317, opened 2024-06-12
refactor(macros): ui tests only for fails

vobradovich:
- left tests only for fails
- move services in mods
- remove `fluent-asserter` crate, use `#[should_panic]`
- add `mutex` in sails-rtl for tests
- reduce test run time from 7:40 to 5:30


rtl/src/gstd/services.rs
```diff
@@ -1,6 +1,34 @@
 use crate::{collections::BTreeMap, MessageId, Vec};
+use core::ops::DerefMut;
 
-static mut MESSAGE_ID_TO_SERVICE_ROUTE: BTreeMap<MessageId, Vec<&'static [u8]>> = BTreeMap::new();
+#[cfg(not(target_arch = "wasm32"))]
+mod map {
+    use super::*;
+    use spin::Mutex;
+
+    static MESSAGE_ID_TO_SERVICE_ROUTE: Mutex<BTreeMap<MessageId, Vec<&'static [u8]>>> =
```
DennisInSky, line 9: nit: Will it look better if we hide static inside a function and remove the mod

macros/tests/integration_tests.rs
DennisInSky, line 1: Should we rename this into something like `ui_error_tests.rs`?

macros/tests/program_basic/mod.rs
DennisInSky, line 1: -> gprogram_basic ?
so it matches with mods for gservice

DennisInSky (approved):
Nice one

DennisInSky (approved):

Issue #320, opened 2024-06-14
client-gen(rust): Change model for query calls


DennisInSky:
### Proposal

Query should differ from command. The latter models sending a message and awating response to that message, whereas the former is just sending a call and receiving response to it as a result of the sent call.

PR #324, opened 2024-06-20
feat(rtl): Add model for query calls

vobradovich:
Add Query trait and impls for `gstd` and `gtest` 

Relates to #320  .


rtl/src/gtest/calls.rs
```diff
@@ -114,4 +114,21 @@ impl Remoting<GTestArgs> for GTestRemoting {
...
+        value: ValueUnit,
+        args: GTestArgs,
+    ) -> Result<Vec<u8>> {
+        let program = self
+            .system
+            .get_program(target.as_ref())
+            .ok_or(RtlError::ProgramIsNotFound)?;
+        let actor_id = args.actor_id.ok_or(RtlError::ActorIsNotSet)?;
+        let run_result =
+            program.send_bytes_with_value(actor_id.as_ref(), payload.as_ref().to_vec(), value);
```
DennisInSky, line 131: Does it make sense to move this code out into some private method?

DennisInSky: Wondering if there is anything special in gtest similar to gclient? Asked @techraed

rtl/src/gstd/calls.rs
```diff
@@ -78,4 +78,31 @@ impl Remoting<GStdArgs> for GStdRemoting {
...
+                args.reply_deposit.unwrap_or_default(),
+            )?
+        } else {
+            msg::send_bytes_for_reply(
+                target,
+                payload,
+                value,
+                args.reply_deposit.unwrap_or_default(),
+            )?
+        };
```
DennisInSky, line 104: Does it make sense to move this code out into some private method and re-use it in both places?

rtl/src/calls.rs
```diff
@@ -172,6 +185,7 @@ impl<TRemoting, TArgs, TReply> Action<TArgs> for RemotingAction<TRemoting, TArgs
 impl<TRemoting, TArgs, TReply> Call<TArgs, TReply> for RemotingAction<TRemoting, TArgs, TReply>
 where
     TRemoting: Remoting<TArgs>,
+    TArgs: Default,
```
DennisInSky, line 188: Why do we need this contraint in these impls if it is not used?

rtl/src/gtest/calls.rs
```diff
@@ -72,6 +72,21 @@ impl GTestRemoting {
         }
         Ok(reply.payload().to_vec())
     }
+
+    fn send_and_get_result(
```
DennisInSky, line 76: Very nit
-> calculate_reply/claclulate_result (similar to how it is called in gear crates)

DennisInSky (approved):

PR #337, opened 2024-06-25
feat(client-gen): Use io module in client calls

### LINKED ISSUE #331: client-gen: Incorporate io module in client code
It would be nice to have this functionality incorporated into the client code so we have a single source of truth

_Originally posted by @DennisInSky in https://github.com/gear-tech/sails/issues/329#issuecomment-2182861070_

holykol:
Resolves #331


DennisInSky:
I suggest closing this PR as solution is not good enough. Let's leave this issue for the time being

Issue #316, opened 2024-06-11
docs: Documentation


DennisInSky:
### Problem to Solve

This issue for collecting various notes about documentation

### Possible Solution

This issue for collecting various notes about documentation

DennisInSky:
There were some ideas in #307

PR #345, opened 2024-06-28
docs: Rework examples

DennisInSky:
Towards #316


holykol (commented):
I would leave at least one example (e.g. ping) as self contained (i.e. have both svc and wasm part) so that it can can be copied to effortlessly start a fresh project. Or we can make 'sails-starter-template' repo so people can fork it.

vobradovich (approved):

PR #349, opened 2024-07-01
feat(idl): Support H160 type

### LINKED ISSUE #346: macros, idl-gen, client-gen: Support H160 as a built-in type
### Problem to Solve

Support H160 as built-in type

### Possible Solution

Similar to the other supported built-ins

### Notes

_No response_

vobradovich:
Resolves #346  .


idl-gen/src/type_names.rs
```diff
@@ -775,6 +719,46 @@ impl TypeName for PrimitiveTypeName {
...
+                    }
+                }
+            }
+        }
+    };
+}
+
+impl_primitive_alias_type_name!(H160, h160);
+impl_primitive_alias_type_name!(H256, h256);
+impl_primitive_alias_type_name!(U256, u256);
```
DennisInSky, line 760: Seems like we can use the same for `ActorId`, `CodeId`, `MessageId`

DennisInSky (approved):
Lovely

DennisInSky (approved):

PR #365, opened 2024-07-05
feat(rtl,client-gen): Support event subscription for GSdk

### LINKED ISSUE #300: rtl, client-gen: Support event subscription for GSdk
### Problem to Solve

At the moment there is no option to subscribe to events emitted by program

### Possible Solution

Introduce some trait for receiving events as bytes and implement it for GTestRemoting. Generate some event listener on a per service basis which allows to receive a typed event

### Notes

_No response_

vobradovich:
Resolves #300 .


DennisInSky (approved):
Wicked!

PR #366, opened 2024-07-05
docs: Add some explanation wrt statics and async

DennisInSky:
Towards #316  .

### DIFF
--- a/examples/rmrk/resource/app/src/services/mod.rs
+++ b/examples/rmrk/resource/app/src/services/mod.rs
@@ -100,11 +100,13 @@ where
             .get_mut(&resource_id)
             .ok_or(Error::ResourceNotFound)?;
 
-        // Caution: If we change the `resource` variable here, the changes will be available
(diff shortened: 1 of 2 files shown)


examples/rmrk/resource/app/src/services/mod.rs
```diff
@@ -100,8 +100,19 @@ where
             .get_mut(&resource_id)
             .ok_or(Error::ResourceNotFound)?;
 
+        // Caution: If we change the `resource` variable here, the changes will be avaialble
```
LouiseMedova, line 103:
```suggestion
        // Caution: If we change the `resource` variable here, the changes will be available
```

LouiseMedova, line 103: May be:
```suggestion
        // Caution: At this point, the execution of `add_part_to_resource` pauses, and all changes made to the state are saved. If we change the `resource` variable here, the changes will be available
```

LouiseMedova (approved):

PR #381, opened 2024-07-12
build: Improve IDL build UX

DennisInSky:
This PR improves IDL build UX. It introduces a new function allowing to generate IDL to the specified path with a check whether IDL file overwritting is actually required.
Also, it disables feature tracking check during WASM build in hope it will minimize build time

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -40,13 +40,13 @@ jobs:
           sudo cp binaryen-$BINARYEN_VERSION/bin/wasm-opt /usr/bin/
 
       - name: Check Code Formatting
-        run: cargo fmt --all --check
+        run: __GEAR_WASM_BUILDER_NO_FEATURES_TRACKING=1 cargo fmt --all --check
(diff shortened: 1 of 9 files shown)


.github/workflows/ci.yml
```diff
@@ -40,13 +40,13 @@ jobs:
           sudo cp binaryen-$BINARYEN_VERSION/bin/wasm-opt /usr/bin/
 
       - name: Check Code Formatting
-        run: cargo fmt --all --check
+        run: __GEAR_WASM_BUILDER_NO_FEATURES_TRACKING=1 cargo fmt --all --check
```
vobradovich, line 43: what's that for? __GEAR_WASM_BUILDER_NO_FEATURES_TRACKING

DennisInSky: When unset this wasm builder forces build script to get re-run on every next build run. Done for the sake of tracking runs with different feature set and propagate it down to wasm building process

PR #385, opened 2024-07-15
feat(sails): make remoting responsible for default args

vobradovich:
Resolves # .


vobradovich:
Still haven't decided on the responsibility for Args - Args::default() vs remoting.args()

PR #392, opened 2024-07-16
fix(workspace): resolve publishing issues

breathx:


Cargo.toml
```diff
@@ -27,18 +28,34 @@ members = [
...
+gear-core-errors = "=1.4.2"
+gprimitives = { version = "=1.4.2", features = ["codec"] }
+gstd = "=1.4.2"
+gtest = "=1.4.2"
+gwasm-builder = { package = "gear-wasm-builder", version = "=1.4.2" }
+
+# Shared with gear deps
+parity-scale-codec = { version = "3.6", default-features = false }
+primitive-types = { version = "0.12", default-features = false }
+scale-info = { version = "2.11", default-features = false }
```
DennisInSky, line 50: I'm not quite convinced with the changed order - it is difficult to decide where to put/find a dependency. You say `Shared with gear` - I'm pretty sure there are plenty of other deps which are shared with gear, but you highlight just some of them due to some reason. Alphabetical order is much more convinient and easy to use

sails/macros/Cargo.toml
```diff
@@ -15,7 +17,7 @@ proc-macro-error.workspace = true
 
 [dev-dependencies]
 parity-scale-codec = { workspace = true, features = ["derive"] }
-sails.workspace = true
+sails = { path = "..", package = "sails-rs" }
```
DennisInSky, line 20: Why?

Cargo.toml
breathx: Gear deps are the most important ones. I hard pinned it to 1.4.2 version, so it's better to keep somewhere together. This section is for gear protocol libraries only (you can mean from "gear-tech/gear" source.

Speaking of "shared" you're totally right. My motivation was to pin the same codec and scale info version as in gear libs to avoid hypothetical collisions but decided to deal with it some time soon (meanwhile cargo kindly manages versions around gear's pinned, so its okay). If you want to remove this section feel free to do it after merging this in the most comfortable for you way. But I'd insist of separation local crates and gear deps as dedicated sections

sails/macros/Cargo.toml
breathx: has no effect overall on testing, debugging, developing, but makes possible to publish: there's cargo and crates-io things with dependencies of publishing crates. It makes it really easier

DennisInSky (approved):

PR #397, opened 2024-07-18
feat(sails, lient-gen): Support mocking generated client

### LINKED ISSUE #92: client-gen(rust): Support mocking generated client
### Problem to Solve

It is not possible to properly mock generated client trait so it can return some results w/o making actual call to the target program

### Possible Solution

1. Turn the `Call` struct into a trait
2. Support one of the crates for mocking so it doesn't require implementing all methods. Possible candidates in order of preferences: [mockall](https://docs.rs/mockall/latest/mockall/), [unimock](https://docs.rs/unimock/latest/unimock/)

### Notes

_No response_

vobradovich:
Resolves #92


rs/Cargo.toml
```diff
@@ -16,13 +16,17 @@ gprimitives.workspace = true
...
+mockall = { workspace = true, optional = true }
 parity-scale-codec = { workspace = true, features = ["derive"] }
 primitive-types.workspace = true
 sails-macros.workspace = true
 scale-info = { workspace = true, features = ["derive"] }
 spin.workspace = true
 thiserror-no-std.workspace = true
 
+[features]
+mockall = ["dep:mockall"]
```
DennisInSky, line 28: Let's have the features table going after all dependencies

rs/client-gen/tests/generator.rs
```diff
@@ -130,5 +130,5 @@ fn test_events_works() {
 fn gen(program: &str, service_name: &str) -> String {
     let program = sails_idl_parser::ast::parse_idl(program).expect("parse IDL");
 
-    sails_client_gen::generate(program, service_name).expect("generate client")
+    sails_client_gen::generate(program, service_name, Some("mockall")).expect("generate client")
```
DennisInSky, line 133: Let's use something similar to the `with_mocks` feature

rs/src/lib.rs
```diff
@@ -16,6 +20,10 @@ pub mod gtest;
 pub mod prelude;
 mod types;
```
DennisInSky, line 22: There is no empty lines between mod imports

DennisInSky (approved):
Wicked

PR #421, opened 2024-07-25
feat(macros): Allow returning non-static lifetime references from service fns #268

### LINKED ISSUE #268: feat: allow returning non-static lifetime references from service fns
The `gservice` macro should support the below code for service methods:
```rust
use core::ptr;

struct Service {
    value: &'static mut u32,
}

impl Service {
    pub fn add<'a>(&'a mut self, v: u32) -> &'a u32 {
        *self.value += v;
        self.value
    }
}

static mut SV: u32 = 12;

fn main() {
    let mut d = Service { value: unsafe { &mut *ptr::addr_of_mut!(SV) } };
    let r = d.add(15);
    print(r);
}

fn print(s: &u32) {
    println!("Value: {}", s);
}
```

vobradovich:
Resolves #268


rs/macros/core/src/service/mod.rs
```diff
@@ -529,7 +529,7 @@ impl<'a> HandlerGenerator<'a> {
     }
 
     fn result_type(&self) -> Type {
-        self.handler.result().clone()
+        type_to_meta_type_with_static_lifetime(self.handler.result().clone())
```
DennisInSky, line 532: shared::func_name
is more ideomatic :-)

DennisInSky: can we rename it into something like
`replace_any_lifetime_with_static_one`
Even though it is long, but it is shorter then the current name and more descriptive to my POV

rs/macros/core/tests/gservice.rs
```diff
@@ -127,3 +127,41 @@ fn works_with_extends_and_lifetimes() {
...
+                }
+            }
+
+            pub async fn first_byte<'a>(&self) -> Option<&'a u8> {
+                unsafe { BYTES.first() }
+            }
+
+            pub async fn last_byte<'a>(&self) -> Option<&'a u8> {
+                unsafe { BYTES.last() }
+            }
```
DennisInSky, line 159: There are too many similar constructions. Having just 
`&'static`
`&'a`
`(&'a, &'static)`
`Option<&'static>`
`Option<&'a>`
seems enough. Impls are not required at all can be just some sort of panic

examples/demo/app/src/references/mod.rs
```diff
@@ -24,12 +24,27 @@ impl ReferenceService {
...
         }
     }
+
+    pub async fn first_byte<'a>(&self) -> Option<&'a u8> {
+        unsafe { BYTES.first() }
+    }
+
+    pub async fn last_byte<'a>(&self) -> Option<&'a u8> {
+        unsafe { BYTES.last() }
+    }
```
DennisInSky, line 47: Seems similar to the prev one

examples/demo/app/src/references/mod.rs
```diff
@@ -24,12 +24,27 @@ impl ReferenceService {
         }
     }
 
+    pub fn add<'a>(&mut self, v: u32) -> &'a u32 {
```
DennisInSky, line 27: Would it be more illustrative if:
1. We make the `ReferenceCount` implementing the `Default` trait
2. Pass it from the outside as `&'a mut ReferenceCount`
3. Hold it as a static in program itself

DennisInSky (approved):

PR #426, opened 2024-07-27
fix: rename dependency cause error

### LINKED ISSUE #401: Provide a way for defining sails crate name in generated code
### Problem to Solve

If the `sails_rs` crate is imported with a different name, the code generated by the macros doesn't compile

### Possible Solution

Add some parameter to the `program` and `service` macros

### Notes

_No response_

ridesun:
Resolves #401  .

I've made some attempts and found that to get the aliases of imported dependencies in Cargo.toml, we need to use the [cargo_metadata](https://github.com/oli-obk/cargo_metadata/) crate to obtain the alias information of dependencies from the project metadata of Cargo. Can we import this new crate?

for example:

```toml
#In Cargo.toml
rename_sails={package = "sails-rs"}
```
```rust
#scale_sails_path::gstd::msg::id() 
//Will be generated to 
rename_sails::gstd::msg::id()
```

@reviewer-or-team


DennisInSky:
@ridesun thanks for the idea. We have already implemented our own solution https://github.com/gear-tech/sails/pull/427. This solution require manual actions, but it covers more possible cases like if one decides to re-export the `sails-rs` crate. Nevertheless, your solution is less demanding in terms of what a person needs to do if they just have a dependency on renamed `sails-rs` crate as well as it doesn't conflict with the solution provided by us. So if you want to go on with your solution, please update it from the `dev` branch

Cargo.toml
```diff
@@ -74,3 +74,4 @@ thiserror = "1.0"
 thiserror-no-std = "2.0"
 tokio = "1.39"
 trybuild = "1"
+cargo_metadata = "0.18"
```
DennisInSky, line 77: Please see the comment saying it has to be in alphabetical order. We also prefer crates to be named in kebab-case instead of snake one

rs/macros/core/Cargo.toml
```diff
@@ -16,6 +16,7 @@ proc-macro-error.workspace = true
 proc-macro2.workspace = true
 quote.workspace = true
 syn = { workspace = true, features = ["full", "extra-traits"] }
+cargo_metadata.workspace = true
```
DennisInSky, line 19: All dependencies are in alphabetical order

ridesun:
Thank you very much for your response. The solution you proposed is very ingenious and covers more possible cases. I will try to continue refining this solution in a new branch according to your requirements. The cargo_metadata crate may have more applications in this project.

PR #427, opened 2024-07-29
feat(macros): Provide a way for defining sails crate name in generated code

vobradovich:
Add to sails macros `crate` property to override `sails_rs` path

Resolves # 401.

### DIFF
--- a/examples/proxy/Cargo.toml
+++ b/examples/proxy/Cargo.toml
@@ -14,7 +14,6 @@ required-features = ["idl-gen"]
 [dependencies]
 demo-client = { path = "../demo/client" }
 sails-idl-gen = { workspace = true, optional = true }
-#sails-rs.workspace = true
 sails-rename = { package = "sails-rs", path = "../../rs"}
 
 [dev-dependencies]
--- a/examples/proxy/Cargo.toml
+++ b/examples/proxy/Cargo.toml
@@ -14,7 +14,8 @@ required-features = ["idl-gen"]
 [dependencies]
 demo-client = { path = "../demo/client" }
 sails-idl-gen = { workspace = true, optional = true }
-sails-rs.workspace = true
+#sails-rs.workspace = true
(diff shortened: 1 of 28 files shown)


rs/macros/core/src/program/args.rs
```diff
@@ -4,13 +4,16 @@ use std::collections::BTreeSet;
 use syn::{
     parse::{Parse, ParseStream},
     punctuated::Punctuated,
-    Ident, Path, Token,
+    Path, Token,
 };
```
DennisInSky, line 9: Usually we don't have empty lines between `use`

rs/macros/core/src/sails_paths.rs
```diff
@@ -1,17 +1,22 @@
...
+use syn::parse_quote;
 
 const SAILS: &str = "sails_rs";
-const SAILS_SCALE_CODEC: &str = concatcp!(SAILS, "::scale_codec");
-const SAILS_SCALE_INFO: &str = concatcp!(SAILS, "::scale_info");
 
-pub(crate) fn scale_types_path() -> syn::Path {
-    syn::parse_str(SAILS).unwrap()
-}
+pub(crate) trait SailsPath {
```
DennisInSky, line 5: Is there an option to go without trait? Doesn't seem it makes much sense

examples/proxy/Cargo.toml
```diff
@@ -14,7 +14,7 @@ required-features = ["idl-gen"]
 [dependencies]
 demo-client = { path = "../demo/client" }
 sails-idl-gen = { workspace = true, optional = true }
-sails-rs.workspace = true
+sails-rename = { package = "sails-rs", path = "../../rs"}
```
DennisInSky, line 17: Better to have it in ws toml, so they are all in the same place

DennisInSky (approved):
Nice one

PR #435, opened 2024-07-30
feat(client-gen): Provide a way for defining sails crate name

### LINKED ISSUE #434: Provide a way for defining sails crate name in `client-gen` generated code
### Problem to Solve

`client-gen` generates code that depends on the path to sails crate `sails_rs::`

### Possible Solution

Add parameter to the `generate` method to override path

### Notes

_No response_

vobradovich:
Resolves #434  .


rs/client-gen/src/lib.rs
```diff
@@ -7,58 +13,106 @@ mod root_generator;
...
+
+    pub fn with_sails_crate(self, sails_path: &'a str) -> Self {
+        Self {
+            sails_path: Some(sails_path),
+            ..self
+        }
+    }
+}
 
-pub fn generate_client_from_idl(
```
DennisInSky, line 17: We want to keep either this or renamed function as a shortcut so people don't need to work with the generator directly when they don't need any of optional params

DennisInSky (approved):

rs/client-gen/src/lib.rs
vobradovich: Leave it as it is for now

PR #438, opened 2024-08-02
feat(sails-rs): Introduce common support for 'with_gas_limit' option in the 'Action' trait

### LINKED ISSUE #436: Introduce common support for 'with_gas_limit' option in the 'Action' trait
At the moment there is a independent support for the `with_gas_limit` option for GSdk and GStd. Recently, GTest was enriched by the same option which means it can be turned into a common one in Sails.
We also want to change the behaviour of this option when running in GStd - instead of doing nothing when `None` is passed, we want to calculate gas under the hood and use the resulting value as gas limit

vobradovich:
GTest calls with gas_limit in PR https://github.com/gear-tech/sails/pull/439

Part of #436 .

### DIFF
--- a/rs/src/calls.rs
+++ b/rs/src/calls.rs
@@ -7,12 +7,12 @@ use core::{future::Future, marker::PhantomData};
 pub trait Action {
     type Args;
 
+    fn with_gas_limit(self, gas_limit: GasUnit) -> Self;
     fn with_value(self, value: ValueUnit) -> Self;
-
     fn with_args(self, args: Self::Args) -> Self;
 
+    fn gas_limit(&self) -> Option<GasUnit>;
     fn value(&self) -> ValueUnit;
-
     fn args(&self) -> &Self::Args;
 }
 
@@ -124,6 +124,7 @@ pub trait Remoting {
         code_id: CodeId,
         salt: impl AsRef<[u8]>,
         payload: impl AsRef<[u8]>,
+        gas_limit: Option<GasUnit>,
         value: ValueUnit,
         args: Self::Args,
     ) -> Result<impl Future<Output = Result<(ActorId, Vec<u8>)>>>;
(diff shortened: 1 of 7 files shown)


rs/src/calls.rs
```diff
@@ -7,12 +7,12 @@ use core::{future::Future, marker::PhantomData};
 pub trait Action {
     type Args;
 
+    fn with_gas_limit(self, gas_limit: GasUnit) -> Self;
```
DennisInSky, line 10: Should it be `Option<GasUnit>` as the `gas_limit` returns an option? Plus otherwise I won't be able to reset it back to initial/default value

rs/src/gstd/calls.rs
```diff
@@ -67,6 +58,7 @@ impl Remoting for GStdRemoting {
         code_id: CodeId,
         salt: impl AsRef<[u8]>,
         payload: impl AsRef<[u8]>,
+        _gas_limit: Option<GasUnit>,
```
DennisInSky, line 61: `create_program_bytes_with_gas_for_reply` should be used when gas is specified

rs/src/gtest/calls.rs
```diff
@@ -125,6 +125,7 @@ impl Remoting for GTestRemoting {
         code_id: CodeId,
         salt: impl AsRef<[u8]>,
         payload: impl AsRef<[u8]>,
+        _gas_limit: Option<GasUnit>,
```
DennisInSky, line 128: Either raise a separate issue for gtest or remove `resolves #...` from this one, maybe replacing it with `towards #...`

rs/src/calls.rs
vobradovich: Leave it as it is

DennisInSky (approved):

PR #449, opened 2024-08-08
feat: introduce Cargo extension CLI for creating packages for building programs

### LINKED ISSUE #448: Introduce CLI tooling for creating Cargo package for programs
### Problem to Solve

ff

### Possible Solution

ff

### Notes

_No response_

DennisInSky:
This PR introduces a binary crate which can be registered as a Cargo extension with the following command line

```
Usage: cargo-sails new-program [OPTIONS] <PATH>

Arguments:
  <PATH>  Path to the new program

Options:
  -n, --name <NAME>  Name of the new program
      --with-client  Generate client crate alongside the program
      --with-gtest   Generate program tests using 'gtest'. Implies '--with-client'
  -h, --help         Print help
```

The `--with-client` and `--with-gtest` options are not implemented yet and generated program contains some stubs for both client and gtest testing

Resolves #448  .


rs/cli/src/main.rs
```diff
@@ -0,0 +1,43 @@
+use clap::Parser;
+use sails_cli::program::ProgramGenerator;
+
+#[derive(Parser)]
+enum Commands {
+    #[command(name = "new-program")]
```
vobradovich, line 6: `new` or `init` shorter

DennisInSky: I've been bearing in mind something like `new-client` so they will be distinguishable

PR #516, opened 2024-09-05
Avoid generating async methods when there are no ones in service impl

### LINKED ISSUE #63: macros: Avoid generating async methods when there are no ones in service impl
### Problem to Solve

The `gservice` macro always generates an async method for processing requests

### Possible Solution

The `gservice` macro should determine if there are no async calls in service impl and generate sync method for processing requests

### Notes

_No response_

ridesun:
Resolves #63  .

I made some changes so that when there are absolutely no async calls in the service impl, only sync methods will be generated.
-
-
-

@reviewer-or-team

### DIFF
--- a/templates/set-vars.rhai
+++ b/templates/set-vars.rhai
@@ -32,7 +32,7 @@ variable::set("mocks-feature-name", "mocks");
 
 // Set versions of used crates
 variable::set("mockall-version", "0.12");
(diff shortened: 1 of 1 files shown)


DennisInSky (changes requested):
This functionaliy requires changes in code generation for program as well.

ridesun:
Noticed this issue and is revising

PR #519, opened 2024-09-06
feat(template): Add query method to template

vobradovich:


templates/program/tests/gtest.rs
```diff
@@ -15,18 +15,45 @@ async fn do_something_works() {
...
+        .await
+        .unwrap();
+
+    let service_client = {{ client_crate_name }}::{{ service-name }}::new(remoting.clone());
+
+    let result = service_client
+        .get_something() // Call service's query (see app/src/lib.rs:19)
+        .recv(program_id)
+        .await
+        .unwrap();
```
breathx, line 56: It’s better to have expects everywhere with explanation of panic

breathx (approved):

templates/program/tests/gtest.rs
vobradovich: fix it in next PR

PR #522, opened 2024-09-11
feat(sails): Allows to return value with reply from a service

### LINKED ISSUE #475: How to reply with value
### Problem

I'd like to make a bank program with a withdraw method which returns previously deposited fund to the caller.

In gstd I was able to implement it by specifying a nonzero value in `msg::reply(payload, value)`

Sails doesn't seem to allow a custom value to be returned with the result, or am I missing something?

### Steps

-

### Possible Solution

_No response_

### Notes

_No response_

### Relevant Log Output

<details><summary>Click to expand/collapse</summary>
<p>

```
<Paste your log here>
```

</p>
</details>

vobradovich:
Resolves #475  .

Introduces special case for handling a `CommandReply<T>` as a result type and extracting a value

### DIFF
--- a/rs/macros/core/src/program/mod.rs
+++ b/rs/macros/core/src/program/mod.rs
@@ -340,8 +340,8 @@ fn generate_handle<'a>(
                 if #input_ident.starts_with(& #service_route_ident) {
                     let program_ref = unsafe { #program_ident.as_ref() }.expect("Program not initialized");
                     let mut service = program_ref.#service_ctor_ident();
-                    let output = service.handle(&#input_ident[#service_route_ident .len()..]).await;
-                    [#service_route_ident .as_ref(), &output].concat()
+                    let (output, value) = service.handle(&#input_ident[#service_route_ident .len()..]).await;
+                    ([#service_route_ident .as_ref(), &output].concat(), value)
                 }
             )
         });
@@ -370,8 +370,8 @@ fn generate_handle<'a>(
         #[gstd::async_main #async_main_args]
         async fn main() {
             let mut #input_ident: &[u8] = &gstd::msg::load_bytes().expect("Failed to read input");
-            let output: Vec<u8> = #(#invocation_dispatches)else*;
-            gstd::msg::reply_bytes(output, 0).expect("Failed to send output");
+            let (output, value): (Vec<u8>, u128) = #(#invocation_dispatches)else*;
+            gstd::msg::reply_bytes(output, value).expect("Failed to send output");
         }
     )
 }
--- a/rs/macros/core/src/program/mod.rs
+++ b/rs/macros/core/src/program/mod.rs
@@ -370,7 +370,7 @@ fn generate_handle<'a>(
         #[gstd::async_main #async_main_args]
         async fn main() {
             let mut #input_ident: &[u8] = &gstd::msg::load_bytes().expect("Failed to read input");
-            let (output, value): (Vec<u8>, u128) = #(#invocation_dispatches)else*;
+            let (output, value): (Vec<u8>, ValueUnit) = #(#invocation_dispatches)else*;
             gstd::msg::reply_bytes(output, value).expect("Failed to send output");
         }
     )
--- a/rs/macros/core/src/service/mod.rs
+++ b/rs/macros/core/src/service/mod.rs
@@ -558,6 +558,10 @@ impl<'a> HandlerGenerator<'a> {
         self.is_query
     }
 
+    fn reply_with_value(&self) -> bool {
+        self.reply_with_value
+    }
+
     fn params_struct(&self) -> TokenStream {
         let params_struct_ident = self.params_struct_ident();
         let params_struct_members = self.handler.params().iter().map(|item| {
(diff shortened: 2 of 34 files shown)


vobradovich:
Awaiting review. Don't merge!

rs/src/gstd/mod.rs
```diff
@@ -41,3 +41,31 @@ impl ExecContext for GStdExecContext {
         *self.msg_id.get_or_init(msg::id)
     }
 }
+
+pub struct CommandResult<T>(T, ValueUnit);
+
+impl<T> CommandResult<T> {
+    pub fn new(value: T) -> Self {
```
DennisInSky, line 48:
```suggestion
    pub fn new(result: T) -> Self {
```

rs/src/gstd/mod.rs
```diff
@@ -41,3 +41,31 @@ impl ExecContext for GStdExecContext {
...
+        Self(value, 0)
+    }
+
+    pub fn with_value(self, value: ValueUnit) -> Self {
+        Self(self.0, value)
+    }
+
+    pub fn to_tuple(self) -> (T, ValueUnit) {
+        (self.0, self.1)
+    }
```
DennisInSky, line 58: Don't like this. I'd prefer seeing something like
```
pub fn result() -> &T {...}
pub fn value() -> ValueUnit {...}
```

rs/src/gstd/mod.rs
```diff
@@ -41,3 +41,31 @@ impl ExecContext for GStdExecContext {
         *self.msg_id.get_or_init(msg::id)
     }
 }
+
+pub struct CommandResult<T>(T, ValueUnit);
```
DennisInSky, line 45: What do you think about naming it as `CommandOutput` in order to avoid confusion with the `Result` type

rs/macros/core/src/shared.rs
```diff
@@ -223,3 +223,28 @@ fn replace_lifetime_with_static_in_path_args(path_args: PathArguments) -> PathAr
         path_args
     }
 }
+
+/// Check if type is `CommandResult<T>` and extract inner type `T`
+pub(crate) fn extract_result_type_with_value(ty: Type) -> (Type, bool) {
```
DennisInSky, line 228: Why do you need `bool` if you have `Option`. Can't you just have a single func which accepts `Type` and returns `Option<Type>`

DennisInSky (commented):
Current impl doesn't prevent returning `CommandResult` from queries
```
pub fn query(&self) -> CommandResult<u128> {
    (123, 100_500).into()
}
```

rs/macros/core/src/shared.rs
vobradovich: The type is always returned either original or unwrapped. See usage

rs/src/gstd/mod.rs
vobradovich: The type is only used within a macro, whe we need some unused methods?

rs/src/gstd/mod.rs
vobradovich: renamed to `CommandReply`

rs/macros/core/src/shared.rs
vobradovich: Eventually changed signature

breathx:
> Current impl doesn't prevent returning `CommandResult` from queries
> ```
> pub fn query(&self) -> CommandResult<u128> {
>     (123, 100_500).into()
> }
> ```

Is it necessary to forbid?

rs/macros/core/src/shared.rs
```diff
@@ -223,3 +223,48 @@ fn replace_lifetime_with_static_in_path_args(path_args: PathArguments) -> PathAr
...
+                if let Some(GenericArgument::Type(ty)) = args.args.first() {
+                    return Some(ty.clone());
+                }
+            }
+        }
+    }
+    None
+}
+
+/// Extract `T` type from `impl Into<CommandReply<T>>`
```
DennisInSky, line 253: I don't like much cryptography in code. becomes unreadable + adds up to maintenance

rs/macros/core/src/service/mod.rs
```diff
@@ -521,8 +540,8 @@ impl<'a> HandlerGenerator<'a> {
         )
     }
 
-    fn result_type(&self) -> Type {
-        shared::replace_any_lifetime_with_static(self.handler.result().clone())
+    fn result_type(&self) -> &Type {
```
DennisInSky, line 543: I would keep it consistent with other methods returning value rather than ref

rs/macros/core/src/program/mod.rs
```diff
@@ -370,8 +370,8 @@ fn generate_handle<'a>(
         #[gstd::async_main #async_main_args]
         async fn main() {
             let mut #input_ident: &[u8] = &gstd::msg::load_bytes().expect("Failed to read input");
-            let output: Vec<u8> = #(#invocation_dispatches)else*;
-            gstd::msg::reply_bytes(output, 0).expect("Failed to send output");
+            let (output, value): (Vec<u8>, u128) = #(#invocation_dispatches)else*;
```
DennisInSky, line 373: Why not ValueUnit

DennisInSky (approved):

PR #523, opened 2024-09-12
feat(cli): integrate sails-js-cli into sails-cli

osipov-mit:


breathx (approved):

rs/cli/src/main.rs
```diff
@@ -12,6 +12,26 @@ struct CargoCommands {
 enum SailsCommands {
     #[command(name = "sails", subcommand)]
     Sails(Commands),
+    #[command(name = "sails-js", subcommand)]
+    SailsJs(JsCommands),
```
vobradovich, line 16: That won't work.
Only one command `sails` per `cargo-sails` executable.

rs/cli/src/main.rs
```diff
@@ -12,6 +12,26 @@ struct CargoCommands {
...
+}
+
+#[derive(Subcommand)]
+enum JsCommands {
+    #[command(name = "generate-client")]
+    GenerateClient {
```
vobradovich, line 22: it is better to make `js-client` under sails, like
```
cargo sails js-client service.idl
```

rs/cli/src/js.rs
```diff
@@ -0,0 +1,71 @@
...
+            .arg(out_path)
+            .arg("-n")
+            .arg(program_name)
+            .spawn()
+            .expect("Failed to run npx sails-js-cli");
+
+        let status = child.wait().expect("An error occured");
+
+        if !status.success() {
+            panic!("Failed to generate JS client");
```
DennisInSky, line 43: Please instead of using `panic`, `expect`, etc., use Result to return the error s it can printed out by the program in standard way. This comment is applicable to the entire mod

rs/cli/src/js.rs
```diff
@@ -0,0 +1,71 @@
...
+        }
+        Ok(())
+    }
+}
+
+pub fn check_app_installed(app: &str) -> bool {
```
DennisInSky, line 49: Doesn't need to be pub

rs/cli/src/js.rs
```diff
@@ -0,0 +1,71 @@
...
+    Command::new("which")
+        .arg(app)
+        .output()
+        .map(|output| output.status.success())
+        .unwrap_or(false)
+}
+
+pub fn check_node() -> Result<()> {
```
DennisInSky, line 57: Doesn't need to be pub

rs/cli/src/js.rs
```diff
@@ -0,0 +1,71 @@
...
+    if !check_app_installed("node") {
+        panic!("Node.js is not installed. Please install Node.js to continue.")
+    }
+
+    Ok(())
+}
+
+pub fn check_npx() -> Result<()> {
```
DennisInSky, line 65: Doesn't need to be pub

rs/cli/src/main.rs
DennisInSky: +1

rs/cli/src/main.rs
DennisInSky: +1

DennisInSky (changes requested):

PR #547, opened 2024-09-25
feat(idl-gen): Transfer comments documenting service methods into IDL and client

### LINKED ISSUE #490: feat(idl-gen): Transfer comments documenting service methods into IDL
### Problem to Solve

It would be nice if documentaion about service methods in rs-code could be transfered into generated IDL

### Possible Solution

Unfortunatelly, rs comments are stripped out of the token stream used to generate service meta. We can introduce a new procedural macro instead which will contain documentation which needs ne transferred to IDL. The macro should be applicable to service methods only

### Notes

_No response_

vobradovich:
Resolves #490


DennisInSky (approved):
Awesome

DennisInSky (approved):

PR #562, opened 2024-10-04
feat(js/cli) add 'yes' flag

Zewasik:
Resolves # .

- fix path for clean build 
- refactor project generation
- make js example private

@osipov-mit


js/.prettierrc
```diff
@@ -0,0 +1,11 @@
+{
```
osipov-mit, line 1: I don't think that this file is needed since we have eslint config

js/cli/src/generate/index.ts
```diff
@@ -3,16 +3,141 @@ import { Sails } from 'sails-js';
...
+    this.isAutomaricOverride = isAutomaricOverride;
+
+    return this;
+  }
+
+  async build() {
+    const rootPath = this.projectPath[0];
+    const srcPath = path.join(...this.projectPath);
+
+    const libFilePath = this.isProject ? path.join(...this.projectPath) : this.projectPath[0];
```
osipov-mit, line 78: If I got it right this path is not the path to lib.ts file but the path to the directory where lib.ts file is generated. If so, let's rename it to make it clearer

js/cli/src/generate/index.ts
```diff
@@ -3,16 +3,141 @@ import { Sails } from 'sails-js';
...
+    }
+
+    if (!this.isProject) {
+      console.log(`Lib generated at ${libFilePath}`);
+      return;
+    }
+
+    if (!existsSync(srcPath)) {
+      mkdirSync(srcPath, { recursive: true });
+    }
```
osipov-mit, line 107: Looks like it's created [here](https://github.com/gear-tech/sails/pull/562/files#diff-6475382e6a6d535733e40b183f13f7faa098c48741ae1df967cba39f26f875f4R80-R82)

js/.prettierrc
Zewasik: eslint does not provide formatting at the moment. I think it would be a good idea if everyone had the same formatting rules

osipov-mit (approved):

PR #579, opened 2024-10-11
feat(js/cli) add support docs

Zewasik:
Resolves #577 .

@osipov-mit

### DIFF
--- a/js/cli/src/generate/service-gen.ts
+++ b/js/cli/src/generate/service-gen.ts
@@ -80,8 +80,11 @@ export class ServiceGenerator extends BaseGenerator {
 
     for (const { name, params, docs } of this._program.ctor.funcs) {
       const args = this.getArgs(params);
-      formatDocs(docs).forEach(line => this._out.line(line, false));
+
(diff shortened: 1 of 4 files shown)


js/cli/src/generate/service-gen.ts
```diff
@@ -77,8 +78,9 @@ export class ServiceGenerator extends BaseGenerator {
   private generateProgramConstructor() {
     if (!this._program.ctor || this._program.ctor.funcs.length === 0) return;
 
-    for (const { name, params } of this._program.ctor.funcs) {
+    for (const { name, params, docs } of this._program.ctor.funcs) {
       const args = this.getArgs(params);
+      formatDocs(docs).forEach(line => this._out.line(line, false));
```
osipov-mit, line 83: There are 2 method for each ctor generated and I think it would be nice to have this docs for each method

Zewasik: Should it be the same comment for each method?

osipov-mit (approved):

PR #588, opened 2024-10-16
chore(sails-rs): move ProgramMeta and ServiceMeta to new crate

vobradovich:


rs/Cargo.toml
```diff
@@ -19,6 +19,8 @@ hashbrown.workspace = true
 hex.workspace = true
 mockall = { workspace = true, optional = true }
 parity-scale-codec = { workspace = true, features = ["derive"] }
+sails-idl-gen = { workspace = true, optional = true }
+sails-idl-meta = { workspace = true }
```
DennisInSky, line 23:
```suggestion
sails-idl-meta.workspace = true
```

Cargo.toml
```diff
@@ -35,6 +36,7 @@ sails-idl-parser = { path = "rs/idl-parser" }
 sails-client-gen = { path = "rs/client-gen" }
 sails-rs = { path = "rs" }
 sails-idl-gen = { path = "rs/idl-gen" }
+sails-idl-meta = { path = "rs/idl-meta" }
```
DennisInSky, line 39: Packages here are going in the order of dependencies, i.e.,
this has to be put first as it doesn't depend on any sails packages.
The `sails-idl-gen` has to be put before `sails-rs` as the latter now depends on the former

vobradovich: ok, sails-idl-meta going first, before macors

DennisInSky (approved):

PR #594, opened 2024-10-21
feat(.net): add generated Gear Api 1.6.2 for .Net

DennisInSky:


net/src/Substrate.Gear.Api/ReplacePattern.ps1
DennisInSky, line 1: This is one of 2 files worth watching

net/src/Substrate.Gear.Api/UpgradeApi.ps1
DennisInSky, line 1: This is one of 2 files worth watching

PR #608, opened 2024-10-23
chore(js): setup release flow

osipov-mit:


js/CHANGELOG.md
DennisInSky, line 1: Why do you want this to be persisted in SCM and not generated by initiating a release via GH?

.github/workflows/release-js.yml
```diff
@@ -0,0 +1,77 @@
+on:
+  push:
```
DennisInSky, line 2: I thought you wanted to support the `release` trigger. Do you want to release on every merge into master regardless of changes?

js/parser/parser.wasm
DennisInSky, line 1: Shouldn't we publish this binary as an artifact of a RS release?

js/CHANGELOG.md
osipov-mit: It's just more customisable and I see no reason to include all commits not related to the js part

.github/workflows/release-js.yml
osipov-mit: Updated. The release will only be triggered on commits affecting root `package.json` and there is a step that checks if the version has been changed

js/parser/parser.wasm
osipov-mit: Yes! 
I will change the behaviour when it's done

.github/workflows/release-js.yml
```diff
@@ -0,0 +1,79 @@
...
+          echo "RELEASE_NOTES<<EOF" >> $GITHUB_ENV
+          cat release_notes.txt >> $GITHUB_ENV
+          echo "EOF" >> $GITHUB_ENV
+          rm release_notes.txt
+
+      - name: 'Create new tag'
+        if: env.skip != 'true'
+        run: |
+          git tag ${{ env.TAG_NAME }}
+          git push origin ${{ env.TAG_NAME }}
```
DennisInSky, line 57: Don't you want to do at the very last point? Does the next action require the tag to exist?

DennisInSky (approved):

.github/workflows/release-js.yml
osipov-mit: It's actually used in the very next step.

.github/workflows/js-ci.yml
```diff
@@ -0,0 +1,57 @@
...
+  pull_request:
+    types: [opened, synchronize, reopened, labeled]
+    branches: [master]
+    paths:
+      - 'js/**'
+      - 'package.json'
+  workflow_dispatch:
+
+env:
+  BINARYEN_VERSION: version_111
```
DennisInSky, line 13: Doesn't seem needed any longer

.github/workflows/js-ci.yml
```diff
@@ -0,0 +1,57 @@
+name: '[js] CI'
+
+on:
+  pull_request:
+    types: [opened, synchronize, reopened, labeled]
+    branches: [master]
+    paths:
+      - 'js/**'
+      - 'package.json'
+  workflow_dispatch:
```
DennisInSky, line 10: What's the point in this trigger if there is nothing is executed when triggered manually?

.gitignore
```diff
@@ -26,6 +25,7 @@ js/test/demo/
 !.yarn/sdks
 !.yarn/versions
 .npmrc
+js/parser/parser.wasm
```
DennisInSky, line 28: Why do you need this change at all?

.github/workflows/js-release.yml
```diff
@@ -0,0 +1,79 @@
...
+      - name: 'Setup NodeJS 20.x'
+        uses: actions/setup-node@v4
+        with:
+          node-version: 20.x
+
+      - name: 'Get version from package.json'
+        run: |
+          VERSION=$(jq -r '.version' package.json)
+          echo "VERSION=$VERSION" >> $GITHUB_ENV
+          echo "TAG_NAME=Sails-JS/v$VERSION" >> $GITHUB_ENV
```
DennisInSky, line 27: To be consistent with the other tags
```suggestion
          echo "TAG_NAME=js/v$VERSION" >> $GITHUB_ENV
```

.github/workflows/js-release.yml
```diff
@@ -0,0 +1,79 @@
...
+
+      - name: 'Get previous version from npm registry'
+        run: |
+          PREVIOUS_VERSION=$(npm show @js/package version)
+          echo "PREVIOUS_VERSION=$PREVIOUS_VERSION" >> $GITHUB_ENV
+
+      - name: 'Compare versions'
+        run: |
+          if [ ${{ env.VERSION }} == ${{ env.PREVIOUS_VERSION }} ]; then
```
DennisInSky, line 36: It is better off to use step outputs as it is easy to lose setting up some var. For example, now this code is highlighted in VSCode, whereas if it were from steps, it would be highlighted only if you forgot to expose this var as an output

DennisInSky (approved):

PR #610, opened 2024-10-25
feat(net): implement IRemoting via SubstrateClient

DennisInSky:


net/src/Sails.Remoting.Abstractions/IRemotingExtensions.cs
```diff
@@ -1,15 +1,17 @@
...
+using GasUnit = Substrate.NetApi.Model.Types.Primitive.U64;
 using ValueUnit = Substrate.NetApi.Model.Types.Primitive.U128;
 
 namespace Sails.Remoting.Abstractions;
 
 public static class IRemotingExtensions
 {
-    public static Task<(ActorId ProgramId, byte[] EncodedReply)> ActivateAsync(
+    /// <inheritdoc cref="IRemoting.ActivateAsync(CodeId, IReadOnlyCollection{byte}, IReadOnlyCollection{byte}, GasUnit?, ValueUnit, CancellationToken)"/>
+    public static Task<Task<(ActorId ProgramId, byte[] EncodedReply)>> ActivateAsync(
```
vobradovich, line 14: Suggest removing these extension methods and making methods with default values.
```
GasUnit? gasLimit = default,
ValueUnit value = ZeroValue,
CancellationToken cancellationToken = default);
```

net/src/Sails.Remoting.Abstractions/IRemotingProvider.cs
```diff
@@ -0,0 +1,14 @@
...
+
+public interface IRemotingProvider
+{
+    /// <summary>
+    /// Creates an instance implementing the <see cref="IRemoting"/> interface
+    /// with initial account for signing transactions.
+    /// </summary>
+    /// <param name="signingAccount"></param>
+    /// <returns></returns>
+    IRemoting GetRemoting(Account signingAccount);
```
vobradovich, line 13: CreateRemoting? Or GetOrCreate if its cached?

net/src/Sails.Remoting/RemotingViaSubstrateClient.cs
```diff
@@ -1,30 +1,63 @@
...
-    public RemotingViaSubstrateClient(RemotingViaSubstrateClientOptions options)
+    /// <summary>
+    /// Creates an instance implementing the <see cref="IRemoting"/> interface via <see cref="SubstrateClientExt"/>
+    /// with initial account for signing transactions.
+    /// </summary>
+    /// <param name="options"></param>
+    /// <param name="signingAccount"></param>
+    public RemotingViaSubstrateClient(
+        RemotingViaSubstrateClientOptions options,
+        Account signingAccount)
```
vobradovich, line 43: My thoughts
Remoting should be a lightweight wrapper over SubstrateClient and extend it with Account data.
The SubstrateClient itself should be Singleton (if possible),  and its lifecycle should be managed from RemotingProvider.

Then, Remoting will be quickly created from the provider and use a shared SubstrateClient, like `Clone`

net/src/Sails.Remoting/RemotingViaSubstrateClient.cs
```diff
@@ -73,7 +204,22 @@ public Task<byte[]> QueryAsync(
...
+                value,
+                cancellationToken)
+            .ConfigureAwait(false);
+
+        // TODO: Check for reply code
+
+        return replyInfo.EncodedPayload;
     }
 
     private async Task<SubstrateClientExt> GetConnectedNodeClientAsync(CancellationToken cancellationToken)
```
vobradovich, line 225: Better move it to another class (strategy), like `NodeProvider` with  `INodeProvider` interface and method `GetConnectedNodeClientAsync`

net/src/Sails.Remoting.Abstractions/IRemoting.cs
```diff
@@ -2,28 +2,63 @@
...
 namespace Sails.Remoting.Abstractions;
 
 public interface IRemoting
 {
-    Task<(ActorId ProgramId, byte[] EncodedReply)> ActivateAsync(
+    /// <summary>
+    /// Sets account for signing transactions.
+    /// </summary>
+    /// <param name="signingAccount"></param>
+    void SetSigningAccount(Account signingAccount);
```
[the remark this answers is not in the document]
DennisInSky: I've decided to put it here in order to avoid any re-connections. Might not needed here at all. The actual design depends on the usage - I would expect it to be used per-scope where one don't need to change user at all.

vobradovich (approved):

PR #681, opened 2024-11-26
feat(net): add Demo Client tests

vobradovich:


net/tests/Sails.DemoClient.Tests/DemoClientTests.cs
```diff
@@ -0,0 +1,214 @@
...
+            .Default()
+            .WithGasLimit(new GasUnit(0))
+            .ActivateAsync(codeId, BitConverter.GetBytes(Random.NextInt64()), CancellationToken.None);
+        // throws on ReceiveAsync
+        var ex = await Assert.ThrowsAsync<ArgumentException>(() => activate.ReceiveAsync(CancellationToken.None));
+
+        // assert
+        // TODO assert custom exception
+        Assert.NotNull(ex);
+    }
```
DennisInSky, line 73: It would nice to see these tests a bit split. These ones, for example, should be in the already created `DemoFactoryTests` class. For the ones, you can create test classes per service `CounterTests`, `PingPongTests`, etc. They can have a base class for the common arrange

DennisInSky (approved):

net/tests/Sails.DemoClient.Tests/CounterTests.cs
```diff
@@ -0,0 +1,107 @@
...
+            .New(counter: new U32(42), dogPosition: dogPosition)
+            .SendReceiveAsync(codeId, BitConverter.GetBytes(Random.NextInt64()), CancellationToken.None);
+
+        var ex = await Assert.ThrowsAsync<ArgumentException>(() => counterClient.Value()
+            .WithGasLimit(new GasUnit(0))
+            .QueryAsync(programId, CancellationToken.None)
+        );
+
+        // assert
+        Assert.NotNull(ex);
```
DennisInSky, line 105: Haven't you asserted the exception above?

PR #699, opened 2024-12-02
fix(net): query reply code parse

vobradovich:


net/src/Substrate.Gear.Client/SubstrateClientExtExtensions.cs
```diff
@@ -594,11 +594,75 @@ public ReplyInfo ToReplyInfo()
...
     }
+
+    /// <summary>
+    /// Convert JToken (JObject) wtih single property to EnumReplyCode
+    /// </summary>
+    /// <param name="token">JObject with single property</param>
+    /// <returns></returns>
+    /// <exception cref="InvalidOperationException"></exception>
+    /// <exception cref="NotImplementedException"></exception>
+    public static EnumReplyCode DeserializeEnumReplyCode(this JToken? token)
```
DennisInSky, line 608: Do we need them public?

DennisInSky (approved):

PR #720, opened 2024-12-09
chore(net): move sails-net-client-gen to root workspace, add cli command

vobradovich:


DennisInSky (commented):
I would suggest putting this PR off (closing) for now as it will introduce a "circular" dependency between RS and NET parts. At the moment the assumption is that RS is an upstream for NET.

PR #767, opened 2025-01-09
feat(sails): Storage trait, impls, examples

vobradovich:


rs/src/gstd/storage.rs
```diff
@@ -0,0 +1,179 @@
...
+    fn get(&'a self) -> impl Storage<Item = T> + 'a {
+        unsafe { NonNull::new_unchecked(self.get()) }
+    }
+}
+
+#[macro_export]
+macro_rules! static_storage {
+    ($type:ty, $init:expr) => {
+        impl $type {
+            pub(crate) fn storage() -> impl Storage<Item = $type> + 'static {
```
breathx, line 154: I'd create mod here maybe so if we declare multiple storages - they don't conflict in namespaces

rs/src/gstd/mod.rs
```diff
@@ -12,6 +12,7 @@ use core::cell::OnceCell;
 pub mod calls;
 pub mod events;
 pub mod services;
+pub mod storage;
```
breathx, line 15: why is it in gstd?

breathx: or it's like everything to be used in programs?

rs/src/gstd/storage.rs
```diff
@@ -0,0 +1,179 @@
...
+    ops::{Deref, DerefMut},
+    ptr::NonNull,
+};
+
+pub type BoxedStorage<T> = Box<dyn Storage<Item = T>>;
+
+pub trait Storage {
+    type Item;
+
+    fn get(&self) -> &Self::Item;
```
breathx, line 13: what about providing also smth like with_data<T>(FnOnce(&Self::Item) -> T) and with_data_mut(FnOnce(&mut Self::Item) -> T) as a reminder of need to free storage access (for cell, for example)

breathx (approved):

rs/src/gstd/mod.rs
vobradovich: Yes, all the stuff for the program under `gstd`

rs/src/gstd/storage.rs
vobradovich: Instead of `impl type`? This is a possible solution, but it would be better to have all storages generated in one place - module or type, like
```rust
static_storage!(service1::Data, service1::Data(0u128));
static_storage!(service2::Data, service2::Data(0u128));
```
generates
```
mod srorage {
  ...
}
```
or
```rust
static_storage!(
    service1::Data => service1::Data(0u128),
    service1::Data => service1::Data(0u128)
);
```

rs/src/gstd/storage.rs
vobradovich: it's not clear why. The storage should be read in the program at the moment when the service constructor is called (once per message). And it is dropped after.

That is why `Storage` is implemented for `RefMut` and not for `RefCell`.

PR #778, opened 2025-01-20
feat(gclient): impl `at_block` for query at specific block

vobradovich:
Add `QueryAtBlock` trait with impl only for `Query<Args = GClientArgs>`

Usage
```
client.query().at_block({hash}).recv(program_id).await;
```


gshep (approved):
:fire:

DennisInSky (approved):

PR #841, opened 2025-03-12
feat: use channels for event testing instead of custom event listener

vobradovich:


rs/macros/tests/gservice_tests.rs
```diff
@@ -197,23 +195,22 @@ async fn gservice_with_lifetimes_and_events() {
...
+    let result = u32::decode(&mut output).unwrap();
+    assert_eq!(result, 42);
 
-        let result = u32::decode(&mut output).unwrap();
-        assert_eq!(result, 42);
+    assert_eq!(output.len(), 0);
 
-        assert_eq!(output.len(), 0);
-    }
+    drop(exposure); // close sender
```
breathx, line 212: why do we need manually drop this?

rs/src/prelude.rs
```diff
@@ -60,3 +60,11 @@ pub use alloy_sol_types;
...
 #[cfg(feature = "ethexe")]
 pub use keccak_const;
+
+/// Channels, Streams and Futures extensions types
+///
+/// See [`futures::stream`], [`futures::channel::oneshot`]
+pub use futures;
+
+/// 'no_std' Async Channel
+pub use async_channel;
```
breathx, line 70: why not just futures::channel::mpsc ?

breathx (approved):

rs/src/prelude.rs
vobradovich: it requires `std` https://docs.rs/futures-channel/0.3.31/src/futures_channel/lib.rs.html#34

rs/macros/tests/gservice_tests.rs
vobradovich: To close sender and invoke `collect()` on Stream
Otherwise, `collect()` future will not be resolved

rs/src/prelude.rs
vobradovich: removed
channels replaced by vec

breathx (approved):

PR #859, opened 2025-03-26
feat(ethexe): add `encode_reply` as second parameter for methods, add Callback selectors

vobradovich:
1. Mirror-Sails Call
    + Method Signature (4 bytes) - calculated from the sails method with added parameters u128 - value, bool - wait for reply
       example: transfer(to: ActorId, amount: u128) -> u128 -> svc1_transfer(uint128,bool,address,uint128)
    + parameters encoded in Solidity
first parameter uint128 - ignored
second bool - encode reply

2. Mirros-Sails Reply
*if encode reply: false*
    + parameters encoded in Solidity (abi encode)
*if encode reply: true*
    + Callback Signature, calculated from reply_for_ + the sails method, message_id, reply type
       example: reply_for_svc1_transfer(uint256,uint128)
    + parameters encoded in Solidity (abi encode)
         message_id: gstd::msg::id()


rs/ethexe/ethapp_with_events/tests/gtest.rs
```diff
@@ -30,7 +30,7 @@ async fn ethapp_with_events_low_level_works() {
     let program = Program::from_file(&system, WASM_PATH);
 
     let ctor = sails_rs::solidity::selector("default(uint128)");
```
StackOverflowExcept1on, line 32:
```suggestion
    let ctor = sails_rs::solidity::selector("default_constructor(uint128)");
```

StackOverflowExcept1on (approved):

PR #875, opened 2025-04-07
BREAKING CHANGE: rename `notify_on` fn to `emit_event`, add `emit_event` fn to Exposure

vobradovich:
Resolves # 872.


breathx (approved):
to consider: `emit_event` -> `emit` but anyway looks good

PR #883, opened 2025-04-11
build(deps): bump up Gear crates to 1.8.0, fix tests, cargo update

vobradovich:


rust-toolchain.toml
```diff
@@ -1,5 +1,5 @@
 [toolchain]
 channel = "stable"
-targets = ["wasm32-unknown-unknown"]
+targets = ["wasm32-unknown-unknown", "wasm32v1-none"]
```
StackOverflowExcept1on, line 3: we could build parser to target `wasm32v1-none` too

rust-toolchain.toml
```diff
@@ -1,5 +1,5 @@
...
 profile = "minimal"
 components = ["rustfmt", "clippy", "rust-src"]
```
StackOverflowExcept1on, line 5: `rust-src` is probably no longer needed because we don't use `-Zbuild-std` in `gear-wasm-builder`. so try just deleting everything and it will use `profile = "default"`.
```suggestion
```

examples/demo/app/tests/gclient.rs
```diff
@@ -160,9 +163,9 @@ async fn demo_returns_not_enough_gas_on_activation() {
 
     assert!(matches!(
         activation_result,
-        Err(sails_rs::errors::Error::Rtl(
-            RtlError::ReplyHasErrorString(s)
-        )) if s.as_str() == "Not enough gas to handle program data"
+        Err(sails_rs::errors::Error::Rtl(RtlError::ReplyHasErrorString(
```
breathx, line 166: If sails still parses strings with unwrapping, this should be removed

breathx (approved):

examples/demo/app/tests/gclient.rs
vobradovich: There's a string parsing in gclient.
https://github.com/gear-tech/gear/blob/v1.8.0/gclient/src/api/listener/mod.rs#L204

rust-toolchain.toml
StackOverflowExcept1on: `wasm32v1-none` is `#![no_std]` target. if this is not possible, use current target.

breathx (approved):

rust-toolchain.toml
vobradovich: This will be done in the next minor issue.

PR #897, opened 2025-04-21
feat(sails-rs): make `with_args` accept Fn to chain calls

vobradovich:


rs/src/gclient/calls.rs
```diff
@@ -20,14 +20,18 @@ pub struct GClientArgs {
 }
 
 impl GClientArgs {
-    pub fn with_voucher(mut self, voucher_id: VoucherId, keep_alive: bool) -> Self {
-        self.voucher = Some((voucher_id, keep_alive));
-        self
+    pub fn with_voucher(self, voucher_id: VoucherId, keep_alive: bool) -> Self {
```
breathx, line 23: why isn't these or similar fns work with just mut self to avoid obj recreation?

rs/src/calls.rs
```diff
@@ -10,7 +10,7 @@ pub trait Action {
     #[cfg(not(feature = "ethexe"))]
     fn with_gas_limit(self, gas_limit: GasUnit) -> Self;
     fn with_value(self, value: ValueUnit) -> Self;
-    fn with_args(self, args: Self::Args) -> Self;
+    fn with_args<F: FnOnce(Self::Args) -> Self::Args>(self, args_fn: F) -> Self;
```
breathx, line 13: this could also be with mut ref

breathx (approved):

rs/src/calls.rs
vobradovich: To chain calls. We can accept `self` or `mut self`
But `self` is better here, since don't need `mut remoting` for query

rs/src/gclient/calls.rs
vobradovich: It's a cheap destruct/construct, w/o clones

PR #910, opened 2025-04-30
feat(macros): accept value to program without a message payload

### LINKED ISSUE #909: feat(macros): accept `value` to program without a message payload
### Problem to Solve

Transfer `value` from one program to another w/o message payload

### Possible Solution

Add macro parameter `accept_transfer`
Accept messages with value and w/o payload.
Do not return a reply - there should be an auto-reply.

```rust
#[sails_rs::program(accept_transfer)]
impl SomeProgram { }
```

### Notes

_No response_

vobradovich:
Resolves #909


rs/macros/core/src/program/mod.rs
```diff
@@ -206,9 +206,19 @@ impl ProgramBuilder {
...
+        let accept_transfer = self.program_args.accept_transfer().then(|| {
+            quote! {
+                if #sails_path::gstd::msg::value() > 0 && #sails_path::gstd::msg::size() == 0 {
+                    return;
+                }
+            }
+        });
+
         let main_fn = quote!(
             #[gstd::async_main #async_main_args]
```
breathx, line 218: Shouldn't it be like this?
```suggestion
            #[#sails_path::gstd::async_main #async_main_args]
```

breathx: I'd double check all entries in future

breathx: seems test-able

breathx (approved):

breathx:
what about naming:
at least it should be `accept_transferS` ?

breathx:
mb analogue to sol with its "payable" ?

techraed:
> mb analogue to sol with its "payable" ?

`payable` holds for contract functions. I believe `accept_transfers` is a good name.

rs/macros/core/src/program/mod.rs
vobradovich: It's ok, use sails_rs::gstd in the module.
```
pub mod wasm {
    use super::*;
    use sails_rs::{gstd, hex, prelude::*};
```

vobradovich:
renamed to `payable`

PR #911, opened 2025-05-01
feat(sails-rs): Use stack buffer to optimize events encoding

techraed:
Same as #859

### DIFF
--- a/rs/src/gstd/utils.rs
+++ b/rs/src/gstd/utils.rs
@@ -1,0 +1,69 @@
+use crate::Output;
+use core::mem::{self, MaybeUninit};
+
+#[derive(Debug)]
+pub(crate) struct MaybeUninitBufferAccessError;
+
+/// A writer that writes to a buffer of `MaybeUninit<u8>` bytes
+/// safely using the `parity-scale-codec::Output` impl.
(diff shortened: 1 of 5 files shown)


rs/src/gstd/utils.rs
```diff
@@ -0,0 +1,55 @@
...
+impl<'a> MaybeUninitBufferWriter<'a> {
+    pub(crate) fn new(buffer: &'a mut [MaybeUninit<u8>]) -> Self {
+        Self { buffer, offset: 0 }
+    }
+
+    /// Gives an access to the internal buffer by providing
+    /// it as a param for the given closure.
+    ///
+    /// Returns an error if the buffer is not initialized.
+    pub(crate) fn access_buffer<T>(&self, f: impl FnOnce(&'a [u8]) -> T) -> T {
```
vobradovich, line 20: nit: `with_buffer` to be consistent

vobradovich (approved):

rs/src/gstd/utils.rs
vobradovich: `Returns an error if the buffer is not initialized` - delete, pls

techraed: done

PR #918, opened 2025-05-07
feat(sails-rs,sails-macro): Use stack buffer to optimize reply payloads

### LINKED ISSUE #867: chore: Use `stack_buffer` to decode and encode messages
### File Location(s)

_No response_

### Proposal

Use `stack_buffer` to decode and encode messages

https://github.com/gear-tech/gear/blob/master/stack-buffer/src/lib.rs#L100

For EthEvent https://github.com/gear-tech/sails/pull/859

techraed:
Resolves #867 

1. The PR suggests to provide to `try_handle` method a new function/closure param, which itself accepts as an argument possibly encoded on the stack reply bytes. Reply bytes are encoded as following: `[service_prefix, method_route, method_invocation_result]`. So instead of returning from the bottom up to the `handle` the encoded result of the method invocation, the handler for the encoded result is provided from the top (`handle`) to the bottom (`try_handle`).
2. `InvocationIO::encode_reply` is removed in favor of `InvocationIO::with_optimized_encode`


rs/macros/core/src/service/exposure.rs
```diff
@@ -203,15 +203,20 @@ impl ServiceBuilder<'_> {
...
+                {
+                    return Some(());
                 }
             }
         });
 
+        // todo [sab] add use for exposure trait
         quote! {
-            pub async fn try_handle(&mut self, #input_ident : &[u8]) -> Option<(#sails_path::Vec<u8>, u128)> {
+            pub async fn try_handle(&mut self, #input_ident : &[u8], result_handler: fn(&[u8], u128)) -> Option<(())> {
```
vobradovich, line 217:
```suggestion
            pub async fn try_handle(&mut self, #input_ident : &[u8], result_handler: fn(&[u8], u128)) -> Option<()> {
```

rs/macros/tests/gservice_tests.rs
```diff
@@ -14,6 +14,18 @@ mod gservice_with_multiple_names;
...
 
+// todo [sab] fix it
+/// Same service name is used for all the tests,
+/// because under the hood exposure call context
+/// stores service names in a static map, which is
+/// accessed by different tests in a multi-threaded
+/// environment. This leads to test's failure in case
+/// race condition occurs.
+const SERVICE_NAME: &str = "TestService";
+fn service_route() -> &'static [u8] {
```
vobradovich, line 25: suggest adding const
```
const SERVICE_ROUTE: &[u8] = &[...]
```

vobradovich (approved):

rs/src/gstd/mod.rs
```diff
@@ -122,11 +124,20 @@ pub trait InvocationIo {
...
-        let mut result = Vec::with_capacity(Self::ROUTE.len() + Encode::size_hint(value));
-        result.extend_from_slice(Self::ROUTE);
-        Encode::encode_to(value, &mut result);
-        result
+    fn with_optimized_encode<T: Encode, R>(
+        value: &T,
+        prefix: &[u8],
+        f: impl FnOnce(&[u8]) -> R,
+    ) -> R {
+        let size = prefix.len() + Self::ROUTE.len() + Encode::size_hint(value);
```
vobradovich, line 132: for security it is better to use here `Encode::encoded_size`

breathx (approved):

PR #952, opened 2025-06-04
feat(client): support to send a message for query with gclient

### LINKED ISSUE #944: feat(gtest,gclient): support for `query` program with sending message, instead of `calculate_for_reply`
### File Location(s)

_No response_

### Proposal

Support force real sending messsage for query via Remoting Params

vobradovich:
Resolves #944 for gclient


vobradovich:
At this moment, command & query differ only at the Sails level as an analogy to HTTP Get, Post requests. .
But in the perspective, it is possible to optimize the execution of `query` at gear protocol level, assuming that query does not mutate the program's state.
But there is still a lot to consider and discuss.

breathx (approved):

PR #953, opened 2025-06-05
chore(sails): display slice as hex w/o alloc

### LINKED ISSUE #938: perf(sails-rs): use `const-hex` crate instead of `hex`
### File Location(s)

_No response_

### Proposal

This fn can be optimized.
Ideally, do not allocate String
```
pub fn unknown_input_panic(message: &str, input: &[u8]) -> ! {
    let mut __input = input;
    let input: String = crate::Decode::decode(&mut __input).unwrap_or_else(|_| {
        if input.len() <= 8 {
             format!("0x{}", crate::hex::encode(input))
         } else {
             format!(
                 "0x{}..{}",
                 crate::hex::encode(&input[..4]),
                 crate::hex::encode(&input[input.len() - 4..])
             )
         }
     });
    panic!("{}: {:?}", message, input)
}

```

vobradovich:
Resolves #938 .

Custom Display impl w/o alloc String

Before & after change
```
[ping_pong_low_level_works] Gas:     1184375865, Size: 115278

[ping_pong_low_level_works] Gas:     1179374017, Size: 114348
```

### DIFF
--- a/examples/demo/app/tests/gtest.rs
+++ b/examples/demo/app/tests/gtest.rs
@@ -492,6 +492,44 @@ async fn counter_add_works_via_manual_mode() {
     );
 }
 
+#[test]
+fn counter_add_low_level_works() {
+    let system = System::new();
+    system.init_logger_with_default_filter("gwasm=debug,gtest=info,sails_rs=debug");
+    system.mint_to(ADMIN_ID, 1_000_000_000_000_000);
+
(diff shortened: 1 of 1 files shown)


techraed:
Looking at #954 I came up to a new idea: name all these tests with "bench" suffix, so when finishing #934 they will be handled (moved or something else) accordingly

Issue #942, opened 2025-05-30
feat(macros): method to return `EventEmitter` primitive for service


vobradovich:
### Proposal

Generate method to return `EventEmitter` primitive for service

Emit event via this primitive, accept non mutable `&self`
```rust
impl<T: Event> EventEmitter<T> {
    pub fn emit_event(&self, event: T) {...}
}
```

vobradovich:
Resolved with https://github.com/gear-tech/sails/pull/984

PR #984, opened 2025-06-26
feat: Changing approach of funcs implementations for services

### LINKED ISSUE #975: refactor: consider changing approach of funcs implementations for services
related to #974 and #970 

I'd consider:
1. hiding routes declared under `#[service]` 
2. implementing them (1) only for exposure, since they're actually related to exposed services
3. removal of option to get event emitter from non-exposed service
4. querying gstd remoting from exposed services only: self.remoting() etc

vobradovich:
Resolves #975 , #942 , #970 , #973

- Service methods defined in `impl Service` with `#[sails_rs::service] `are moved to `Exposure` and are no longer accessible directly via the Service type.
- Exposure implement `DerefMut<Target = Service>` to allow transparent access to service internals.
- Event emission (`emit_event`, `emit_eth_event`) is permitted only within Exposure methods. This removes the need for statically storing the service route, simplifying event emission logic.
- `message_id` removed from Exposure, and instead made accessible exclusively through Syscalls.
- Introduced `ExposureWithEvents` trait to handle event emissions for services.
- Simplified event emission methods by utilizing the new `EventEmitter` struct.
- Enhanced async handling in service methods and improved code organization.
- `Exposure` moved to `try_handle` and `try_handle_async` instead of borrowing it
- Base services no longer need to implement `Clone`, an extended service must implement `Into` to destruct into base services
- Implement `Into` trait for exposure to base exposure (in case of single extend)
- BREAKING CHANGE: the `export` attribute must be on all methods that should be accessible via message.
- Fixed various documentation typos and improved clarity in comments.
- Updated tests to reflect changes in event handling and exposure.


rs/src/gstd/events.rs
```diff
@@ -64,26 +65,140 @@ fn with_optimized_event_encode<T, E: SailsEvent, F: FnOnce(&[u8]) -> T>(
...
-where
-    TEvents: SailsEvent,
-{
-    let route = crate::gstd::services::route(crate::gstd::syscalls::Syscall::message_id());
-    __emit_event_with_route(route, event)
+/// An event emitter that allows emitting events of type `T` for a specific route.
+///
+/// This is lightweight and can be cloned.
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct EventEmitter<T> {
```
breathx, line 72: I'd also put the bound for T: Event here since the type is meaningless without it

rs/ethexe/macros-tests/tests/service_with_extends_and_lifetimes/mod.rs
```diff
@@ -41,17 +43,20 @@ impl<'a> ExtendedWithLifetime<'a> {
...
+    #[export]
     pub fn name(&self) -> String {
         NAME_RESULT.to_string()
     }
 }
 
-impl<'a> AsRef<BaseWithLifetime<'a>> for ExtendedWithLifetime<'a> {
-    fn as_ref(&self) -> &BaseWithLifetime<'a> {
-        &self.base
+#[allow(clippy::from_over_into)]
```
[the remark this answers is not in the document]
vobradovich: it is up to the program developer if one prefers to use `derive_more`.

rs/src/gstd/events.rs
vobradovich: Here are two impl with different bounds
```
impl<T: SailsEvent> EventEmitter<T> {

#[cfg(feature = "ethexe")]
impl<T: super::EthEvent> EventEmitter<T> {
```

PR #1016, opened 2025-08-05
ci: introduce benchmarks for CI

techraed:
This PR introduces:
- CI job which runs benchmarks in case `run-benchmarks` label is set in a PR and specific paths were changed (`rs/`, `benchmarks/`, `examples/`, `Cargo.*`, and the job config). Also in case the PR is pushed to `master`, the bench will run.
- `bench-analyzer` scripts which reads 2 files with bench data and prints a table with bench statistics. Optionally returned table can be set to the file. Also in case `threshold` is introduced, if any bench doesn't pass threshold (i.e. current bench differs from the other by threshold value) the script ends with an error


techraed:
github benchamarks action doesn't work with custom benches

adjusting solution to do not working steps on our own

PR #1071, opened 2025-10-03
feat(js): expose payload decoding methods from `TransactionBuilder`

### LINKED ISSUE #991: Add the ability to decode payload/result for a specific function of generated program lib
To handle payload decoding during batch transactions from different programs, would be useful to have a method that will allow to decode replies manually - in the same way it happens in `TransactionBuilder`.

Example of possible usage: https://github.com/gear-tech/gear-bridges/blob/f006cd01f8d86c8579ce2301adab6af1063a7e14/frontend/src/features/swap/hooks/vara/use-handle-vara-submit.ts#L114

nikitayutanov:
Resolves #991.

### DIFF
--- a/js/src/transaction-builder.ts
+++ b/js/src/transaction-builder.ts
@@ -9,7 +9,7 @@ import {
 } from '@gear-js/api';
 import { SignerOptions, SubmittableExtrinsic } from '@polkadot/api/types';
 import { IKeyringPair, ISubmittableResult } from '@polkadot/types/types';
-import { TypeRegistry, u128, u64 } from '@polkadot/types';
+import { Bytes, TypeRegistry, u128, u64 } from '@polkadot/types';
 import { getPayloadMethod } from 'sails-js-util';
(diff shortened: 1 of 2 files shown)


js/src/transaction-builder.ts
```diff
@@ -332,6 +333,27 @@ export class TransactionBuilder<ResponseType> {
...
+    const type = this._registry.createType<any>(this._responseType, noPrefixPayload);
+
+    return type[method]();
+  }
+
+  /**
+   * ## Check if the reply is an error and throw if it is
+   * @param message - UserMessageSent message
+   */
+  public checkErrorReply({ payload, details }: GearCoreMessageUserUserMessage) {
```
osipov-mit, line 353: Well, probably `throwOnErrorReply` method name would be better, or at least `handleErrorReply`, since `check` suggests boolean return

osipov-mit (changes requested):

osipov-mit (approved):

PR #1113, opened 2025-11-21
feat(rs/ethexe): add some logic to demonstrate how vara.eth works

StackOverflowExcept1on:
This is mainly to show that you can calculate something complex on vara.eth and then get the result in EVM contract using reply hooks. I also added `expect` to show how EVM contract can handle panics (error replies).

### DIFF
--- a/rs/ethexe/ethapp_with_events/src/lib.rs
+++ b/rs/ethexe/ethapp_with_events/src/lib.rs
@@ -37,7 +37,7 @@ pub struct SomeService;
 impl SomeService {
     #[sails_rs::export]
     pub async fn do_this(&mut self, p1: u32, p2: sails_rs::String) -> u32 {
-        let r1 = p1 * 2;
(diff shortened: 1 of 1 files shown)


vobradovich (approved):

StackOverflowExcept1on:
@vobradovich Well, these are at least some calculations for demonstration purposes. For example, user creates program that performs multiplication (https://hoodi.etherscan.io/address/0xdb892e7fba8349d4aeddae9521d4e906d8140539#writeContract, https://hoodi.etherscan.io/tx/0xee3a23a444887ba323f8c3015d4149758823336fd570037d21c663f950d829a5) and can then request the result by reading the contract (https://hoodi.etherscan.io/address/0xdb892e7fba8349d4aeddae9521d4e906d8140539#readContract). In this case, `getSvc1DoThisResult(42, "hello")` will return `84`.

PR #1128, opened 2025-12-03
fix: use RustType from SolType to decode payload

vobradovich:
Fixed usage of `ActorId`, `[u8; N]`


StackOverflowExcept1on:
It would also be good idea to look at [`SolType` implementors](https://docs.rs/alloy-sol-types/latest/alloy_sol_types/trait.SolType.html#implementors). Some types may require `calldata` (it is better to use `calldata` instead of `memory` in most cases).

breathx (approved):

ukint-vs (approved):

PR #1134, opened 2025-12-05
feat(idl-gen): type resolver, generic resolver, builder

vobradovich:


techraed (commented):
Awesome simplification of the `idl-gen` was done. That's notable especially when considering struct/enum fields type names resolution.

Some things to consider:
1. I strongly recommend to move tests from #1101 PR from modules `meta2.rs`, `type_names.rs`, `idl-gen/tests/generator.rs`, `macros/tests`. Passing these tests will fully assure that idl-gen works in accordance to the spec introduce in the mentioned PR
2. Interface ids are already included in the `master-idlv2`. So checks for repeated definition of the base services done in the macro by comparing string names can be changed to generating const asserts comparing `INTERFACE_ID`s of base services.
3. `idl-gen` must not generate the service several times. The check for duplicated interface id can be done in `build_program_ast` function

techraed (approved):

PR #1138, opened 2025-12-07
feat(rs/cli/program_new): add dependency inheritance from workspace

StackOverflowExcept1on:
`--app` seems useless and complicates the implementation, so I removed it. Perhaps we could consider `--eth`, `--client`, `--no-client`.

### DIFF
--- a/rs/ethexe/ethapp_with_events/src/lib.rs
+++ b/rs/ethexe/ethapp_with_events/src/lib.rs
@@ -1,5 +1,7 @@
 #![no_std]
 
+#![allow(unused_assignments)]
+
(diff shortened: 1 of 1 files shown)


StackOverflowExcept1on:
@vobradovich I had problem where I needed to add `\n\n` to the `.askama` files because for some reason this template engine ignored just one `\n` so the last line would be empty

vobradovich:
> @vobradovich I had problem where I needed to add `\n\n` to the `.askama` files because for some reason this template engine ignored just one `\n` so the last line would be empty
```
{{- "\n" -}}
```
works on both Windows and Linux

PR #1198, opened 2026-01-29
feat(sails-js): impl IDL v2 parser, AST types, SailsProgram, SailsService classes

vobradovich:


js/parser/rollup.config.js
```diff
@@ -40,7 +40,9 @@ function writeCompressedWasmParser(type) {
     async closeBundle() {
       const base64Bytes = await getBase64Parser(config['sails-rs']);
 
+      mkdirSync('./lib', { recursive: true });
```
osipov-mit: It works perfectly for me after removing these commands.

js/src/sails-v2.ts
```diff
@@ -0,0 +1,513 @@
...
+
+    // this.generateScaleCodeTypes();
+    this._services = this._initServices();
+    this._resolveServiceUnit = (ident: IServiceIdent) => {
+      if (!ident.interface_id) {
+        throw new Error(`Service "${ident.name}" is missing interface_id in IDL`);
+      }
+      const interfaceId = InterfaceId.from(ident.interface_id).asU64();
+      return this._services.get(interfaceId);
+    };
```
osipov-mit: And I don't understand why it's defined in constructor. I think it can be just a normal private method

PR #1248, opened 2026-03-20
feat: replace scale-info with sails-type-registry

m62624:


rs/idl-gen/src/type_resolver.rs
```diff
@@ -532,23 +516,16 @@ mod tests {
...
-            .register_type(&MetaType::new::<gprimitives::H256>())
-            .id;
-        let h256_as_generic_param_id = registry
-            .register_type(&MetaType::new::<GenericStruct<gprimitives::H256>>())
-            .id;
+        let h256_id = registry.register_type::<gprimitives::H256>();
+        let h256_as_generic_param_id = registry.register_type::<GenericStruct<gprimitives::H256>>();
 
-        let portable_registry = PortableRegistry::from(registry);
+        let portable_registry = &registry;
```
vobradovich: remove, pls, all lines

rs/idl-gen/src/type_resolver.rs
```diff
@@ -586,31 +559,25 @@ mod tests {
...
+        let bool_u32_enum_id = registry.register_type::<GenericEnum<bool, u32>>();
+        let portable_registry = &registry;
+        let resolver = TypeResolver::from_registry(portable_registry);
 
         let u32_string_enum = resolver.get(u32_string_enum_id).unwrap();
         assert_eq!(u32_string_enum.to_string(), "GenericEnum<u32, String>");
 
         let bool_u32_enum = resolver.get(bool_u32_enum_id).unwrap();
-        assert_eq!(bool_u32_enum.to_string(), "GenericEnum<bool, u32>");
+        assert_eq!(bool_u32_enum.to_string(), "TestsGenericEnum<bool, u32>");
```
vobradovich: must be `GenericEnum<bool, u32>`, right?

vobradovich: AI explain

For GenericStruct<T> — bare parameter → same TypeRef
The field field: T is a bare type parameter ident, so the macro emits:


registry.register_type_def(Type::builder().name("T").parameter("T"))
This creates a TypeDef::Parameter("T") entry. Crucially, register_type_def deduplicates by name+path: since name = "T" is non-empty, the second call finds the same entry and returns the same TypeRef P for both GenericStruct<u32> and GenericStruct<String>.

So is_fields_equal sees [TypeRef P] for both → they match → both get the name "GenericStruct".

The concrete type (u32 vs String) lives only in type_params, not in the field's TypeRef. The resolver reads type_params separately to fill in generics: [U32] vs generics: [String].

For GenericEnum<T1, T2> — nested generics → different TypeRefs
GenericEnum has Variant4(Option<(T1, GenericStruct<T2>, u32)>). The inner GenericStruct<T2> is not a bare param, so the macro emits:


let base_id = registry.register_type::<GenericStruct<T2>>(); // concrete: GenericStruct<String> or GenericStruct<u32>
let args = vec![TypeRef(Parameter("T2"))];
registry.register_type_def(Type::builder().applied(base_id, args))
Here base_id is the TypeRef of the concrete instantiation — different for GenericStruct<String> vs GenericStruct<u32>. The Applied node doesn't normalize (since GenericStruct isn't Option/Result/Vec/etc.), so the stored TypeRefs differ. This bubbles up: the tuple and Option<tuple> field also differ between instantiations.

So is_fields_equal sees different TypeRefs → collision → unique_type_name walks up the module path → prepends "tests" → "TestsGenericEnum".

rs/idl-gen/src/type_resolver.rs
```diff
@@ -98,121 +96,183 @@ impl<'a> TypeResolver<'a> {
...
-            let type_decl = self.resolve_type_decl(&pt.ty)?;
-            self.map.insert(pt.id, type_decl);
+    fn build_type_decl_map(&mut self, exclude: BTreeSet<TypeRef>) -> Result<()> {
+        let all_types: Vec<_> = self.registry.types().collect();
+        for (id, _) in all_types {
+            if exclude.contains(&id) {
+                continue;
+            }
+            if let Ok(type_decl) = self.resolve_by_id(id) {
+                self.map.insert(id, type_decl);
```
vobradovich: is it suppresses real errors?
`let type_decl = self.resolve_by_id(id)?;`

rs/type-registry/Cargo.toml
```diff
@@ -0,0 +1,23 @@
...
+
+[dev-dependencies]
+insta.workspace = true
+trybuild.workspace = true
+sails-type-registry = { path = ".", features = ["derive", "gprimitives"] }
+
+[features]
+default = []
+derive = ["dep:sails-type-registry-derive"]
+gprimitives = ["dep:gprimitives"]
```
vobradovich, line 22:
```
[dev-dependencies]
insta.workspace = true
trybuild.workspace = true

[features]
default = ["derive"]
derive = ["dep:sails-type-registry-derive"]
gprimitives = ["dep:gprimitives"]

[[test]]
name = "gprimitives"
required-features = ["gprimitives"]
```

rs/type-registry/Cargo.toml
```diff
@@ -0,0 +1,23 @@
...
+rust-version.workspace = true
+
+[dependencies]
+gprimitives = { workspace = true, optional = true }
+sails-type-registry-derive = { path = "derive", optional = true }
+
+[dev-dependencies]
+insta.workspace = true
+trybuild.workspace = true
+sails-type-registry = { path = ".", features = ["derive", "gprimitives"] }
```
vobradovich:
```suggestion
sails-type-registry-derive = { workspace = true, optional = true }
```

PR #1300, opened 2026-04-14
fix: strip .wasm/.opt extensions from program name

m62624:
```bash
cargo install --path rs/cli

cargo sails new ~/tmp/test-bug --name test-bug --sails-path ./rs

cd ~/tmp/test-bug && cargo build --release

grep '^program' ~/tmp/test-bug/target/wasm32-gear/release/test_bug.idl

```

### DIFF
--- a/rs/src/builder.rs
+++ b/rs/src/builder.rs
@@ -96,7 +96,7 @@ impl<P: ProgramMeta> ClientBuilder<P> {
             .to_string_lossy()
             .split('.')
             .next()
-            .unwrap_or("")
+            .expect("path file_name must have at least one segment before a dot")
             .to_string();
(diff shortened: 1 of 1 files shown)


rs/src/builder.rs
```diff
@@ -94,12 +94,18 @@ impl<P: ProgramMeta> ClientBuilder<P> {
             .file_name()
             .expect("Invalid path to wasm")
             .to_string_lossy()
+            .split('.')
+            .next()
+            .unwrap_or("")
```
vobradovich: better `.expect(...)`, not `.unwrap_or("")`

vobradovich (approved):

PR #1313, opened 2026-04-21
feat: add Address type

### LINKED ISSUE #1262: Add `Address([u8; 20])` and replace direct use of `alloy_primitives::Address`
### File Location(s)

_No response_

### Proposal

Introduce a local `Address([u8; 20])` primitive in `gprimitives` and use it instead of `alloy_primitives::Address` in the Ethereum-related conversion layer.

## Problem

Current `ethexe` support in sol_types.rs depends directly on `alloy_primitives::Address`:

- `impl From<Address> for ActorId`
- `impl From<ActorId> for Address`
- `type SolType = <Address as SolValue>::SolType`

This makes the address type external to `gprimitives` and prevents us from having a SCALE-native address primitive with the expected derives/metadata support.

## Proposal

Add a new local type:

```rust
pub struct Address([u8; 20]);
```

and implement/derive:

- `scale::Encode`
- `scale::Decode`
- `scale_info::TypeInfo`
- `alloy_sol_types::SolValue`

Then replace direct usage of `alloy_primitives::Address` in the `ethexe` bridge with this new type, while preserving Solidity ABI compatibility.

## Scope

- Add `Address([u8; 20])`
- Implement conversions between `Address` and `ActorId`
- Keep `SolValue` / ABI encoding and decoding behavior equivalent to the current implementation
- Add tests covering:
  - `Address <-> ActorId` conversion
  - ABI encode/decode compatibility
  - SCALE `Encode` / `Decode` roundtrip
  - `TypeInfo` availability

## Acceptance Criteria

- `Address` supports SCALE `Encode`, `Decode`, and `TypeInfo`
- Existing Ethereum ABI behavior remains unchanged
- Tests validate both SCALE and Solidity compatibility

m62624:
Resolves #1262

### DIFF
--- a/rs/src/lib.rs
+++ b/rs/src/lib.rs
@@ -38,7 +38,11 @@ pub mod gstd;
 pub use gtest;
 #[cfg(all(feature = "mockall", not(target_arch = "wasm32")))]
 pub use mockall;
+#[cfg(feature = "ethexe")]
+mod address;
 pub mod prelude;
+#[cfg(feature = "ethexe")]
+pub use address::Address;
 #[cfg(feature = "ethexe")]
 pub mod solidity;
 mod types;
--- a/rs/src/lib.rs
+++ b/rs/src/lib.rs
@@ -42,8 +42,6 @@ pub use mockall;
 mod address;
 pub mod prelude;
 #[cfg(feature = "ethexe")]
-pub use address::Address;
-#[cfg(feature = "ethexe")]
 pub mod solidity;
 mod types;
 mod utils;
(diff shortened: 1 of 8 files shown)


rs/src/lib.rs
```diff
@@ -38,8 +38,12 @@ pub mod gstd;
 pub use gtest;
 #[cfg(all(feature = "mockall", not(target_arch = "wasm32")))]
 pub use mockall;
+#[cfg(feature = "ethexe")]
+mod address;
 pub mod prelude;
 #[cfg(feature = "ethexe")]
+pub use address::Address;
```
vobradovich: double re-export
sails_rs::Address
sails_rs::prelude::Address;

vobradovich (approved):

PR #1323, opened 2026-04-23
feat(js): service-scoped type resolution + generic substitution on TypeResolver

### LINKED ISSUE #1317: sails-js v2: TypeResolver needs service-scoped lookup + generic substitution for downstream walkers
## Summary

Two related correctness bugs hit any consumer that walks v2 `TypeDecl` to do custom validation, coercion, or codegen:

1. **No service scope for type resolution.** `SailsProgram` keeps user types per-service internally (`_doc.services[i].types`), but there's no public API to resolve `TypeDecl { kind: 'named', name: 'Packet' }` in the context of a specific service. Two services declaring the same-named struct with different shapes silently collide when consumers flatten types into a global map.

2. **No public generic substitution.** `TypeResolver.getTypeDeclString` handles generic substitution internally when rendering a name, but doesn't expose "resolve a TypeDecl to its fully-substituted form" — so consumers walking `TypeDecl` for custom purposes (hex-to-bytes coercion, JSON schema derivation, payload validation) have to re-implement the substitution logic themselves.

Both surfaced while integrating `sails-js@1.0.0-beta.1` into a wallet CLI — see [gear-foundation/vara-wallet#29](https://github.com/gear-foundation/vara-wallet/pull/29) and the workaround write-up at [docs/sails-js-upstream-proposals.md](https://github.com/gear-foundation/vara-wallet/blob/feat/sails-js-1.0.0-beta-dual-idl/docs/sails-js-upstream-proposals.md) (proposals #2 and #3).

## Reproducer 1 — collision

```idl
!@sails: 1.0.0-beta.1

service A@0x... {
    functions { Set(p: Packet); }
    types { struct Packet { payload: [u8; 4], } }
}

service B@0x... {
    functions { Set(p: Packet); }
    types { struct Packet { payload: [u8; 8], } }
}
```

A consumer that builds a `Map<string, Type>` by iterating `_doc.services[i].types` across both services ends up with only one `Packet` entry. Depending on iteration order, `A.Set`'s 4-byte payload is either rejected (wrong length against B's schema) or silently mis-encoded. There's no public API that narrows a type lookup to a single service's scope.

## Reproducer 2 — generics

```idl
service Gen@0x... {
    functions {
        SetPayload(p: Envelope<[u8]>);
    }
    types {
        struct Envelope<T> {
            id: u32,
            payload: T,
        }
    }
}
```

The arg's `typeDef` comes through as `{ kind: 'named', name: 'Envelope', generics: [{ kind: 'slice', item: 'u8' }] }`. When a consumer walks this and recurses into `Envelope`'s fields, `payload`'s type is literally `{ kind: 'named', name: 'T' }` — a bare reference to the type_param. Without substitution, the walker passes hex strings through untouched and downstream SCALE encoding fails or silently produces wrong bytes.

## Current workaround

vara-wallet carries ~180 lines in `src/utils/hex-bytes.ts` and `src/services/sails.ts` implementing:
- a `V2Substitutions` map threaded through the walker recursion
- a `resolveV2Subs` one-level lookup
- a `getRegistryTypes(sails, serviceName?)` helper that narrows to program-level + one service's types

Every consumer that wants to do its own payload validation will reinvent the same logic.

## Proposed API

Expose on `TypeResolver`:

```ts
class TypeResolver {
  // Resolve a TypeDecl to its concrete definition in a service context.
  // Program-level types are always visible; service-local types shadow
  // only within that service's scope.
  resolveInService(serviceName: string, typeDecl: TypeDecl): Type | undefined;

  // Fully substitute type_params through a TypeDecl tree. Idempotent.
  // Consumers can feed the result to their own walkers without
  // worrying about generic substitution.
  substituteGenerics(
    typeDecl: TypeDecl,
    generics?: TypeDecl[],
    paramNames?: string[],
  ): TypeDecl;
}
```

Alternatively, if you prefer a callback-based walker:

```ts
class TypeResolver {
  walkTypeDecl(
    typeDecl: TypeDecl,
    visitor: {
      onPrimitive?(name: PrimitiveType, ctx: WalkerCtx): void;
      onSlice?(item: TypeDecl, ctx: WalkerCtx): void;
      onArray?(item: TypeDecl, len: number, ctx: WalkerCtx): void;
      onTuple?(types: TypeDecl[], ctx: WalkerCtx): void;
      onStruct?(def: ITypeStruct, resolvedFields: IStructField[], ctx: WalkerCtx): void;
      onEnum?(def: ITypeEnum, resolvedVariants: IEnumVariant[], ctx: WalkerCtx): void;
    },
    ctx?: { serviceName?: string; generics?: Map<string, TypeDecl> },
  ): void;
}
```

Either shape would let consumers get service-scoped, fully-substituted `TypeDecl` values without reaching into `_doc` or re-implementing the resolution logic that the library already has internally.

## Willingness to contribute

Happy to open a PR if the design direction is agreed. The consumer-side implementation in vara-wallet (at [src/utils/hex-bytes.ts `coerceHexToBytesV2`](https://github.com/gear-foundation/vara-wallet/blob/feat/sails-js-1.0.0-beta-dual-idl/src/utils/hex-bytes.ts)) is a sketch of the substitution algorithm that could port inward.

ukint-vs:
Closes #1317.

> **Rework (2026-05-06)** — addresses [vobradovich's CHANGES_REQUESTED review](https://github.com/gear-tech/sails/pull/1323#pullrequestreview-4234814306) ("violates self-sufficient service IDL concept"). Aligns this PR with the contract being formalized in #1341: a service IDL is independently distributable; service signatures resolve types from the service's own `types` plus its `extends` chain, never from `program.types` ambient scope. Internal-only API change — `TypeResolver` and the new `resolveNamed`/`resolveGenerics`/`substituteGenerics`/`resolveInService` surface stays the same. See the latest commit for details.

## Summary

Adds three public methods on `TypeResolver` and a convenience accessor on `SailsProgram` so downstream consumers (wallets, explorers, schema tooling, custom codegen) stop re-implementing type-scope and generic-substitution out-of-band.

**New API**

```ts
class TypeResolver {
  // Constructor: a single pre-merged Type[]; later entries override earlier ones
  // on name collision (last-write-wins).
  constructor(types: Type[]);

  // Look up a named user Type from this resolver's scope.
  // With concrete generics, returns a substituted concrete Type with type_params stripped.
  resolveNamed(type: TypeDecl): Type | undefined;
  resolveNamed(name: string, generics?: TypeDecl[]): Type | undefined;

  // Recursively substitute type parameters through a TypeDecl tree.
  // Pure, idempotent, same-ref return on unchanged subtrees.
  // Throws on cyclic substitution maps ({T: T} or {T: U, U: T}).
  substituteGenerics(typeDecl: TypeDecl, substitutions?: Record<string, TypeDecl>): TypeDecl;

  // Zip user type's type_params with a concrete generics list.
  resolveGenerics(userType: Type, generics?: TypeDecl[]): Record<string, TypeDecl>;
}

class SailsProgram {
  // Direct AST lookup scoped to a named service. Resolves from the service's own
  // `types` plus the types of every service it extends, transitively. Program-level
  // `types` are NOT in scope (they belong to program/ctor declarations).
  // Backed by a lazy O(1) Map index — safe to call in tight walker loops.
  resolveInService(serviceName: string, typeDecl: TypeDecl): Type | undefined;
}
```

Per-service `TypeResolver` instances seeded by `SailsProgram.services` see only the service's own `types` plus the transitive `extends` chain. The chain walk is depth-first with cycle detection; on a pathological `A extends B, B extends A` graph the constructor throws with the chain in the message. Service-local types shadow base-service types on name collision.

`getTypeDeclString` was also updated to call the renamed `resolveGenerics` helper instead of the inline copy at lines 229-233.

## Why

Two correctness bugs in consumer code before this change:

1. **Name collisions across services.** Two services each declaring `Packet` with different shapes silently collided when a consumer flattened types into a single map. No public API narrowed a lookup to a single service's scope.
2. **Generic substitution.** `Envelope<[u8]>`'s `payload` field came through as a bare `{kind:'generic', name:'T'}` leaf, causing walkers to silently pass hex strings through untouched or SCALE-encode with wrong bytes. The substitution algorithm already lived inside `getTypeDeclString` — this PR factors it out and exposes it.

vara-wallet had ~180 lines of workaround (`coerceHexToBytesV2`, `getRegistryTypes`, ad-hoc substitution recursion) that can now collapse to a few method calls.

## Tests

24+ tests across the resolver + parser-resolver suites.

**Unit** (`js/test/idl-v2-type-resolver.test.ts`):
- `substituteGenerics`: bare-param substitution, recursion through slice/array/tuple/named-with-generics, pass-through for unknown names, idempotence, non-mutation, chain resolution (`T → U → u32`), self-referential throw, cyclic chain throw, unknown-kind throw.
- `resolveNamed`: returns user `Type` for named decl (raw and concrete-generics paths), returns `undefined` for primitives/slices/arrays/tuples/unknown names/bare type_params.
- `resolveGenerics`: zip, no type_params, extra generics truncated.
- Last-write-wins merge: locals override on name collision; registry reflects the last-declared shape.
- **Cycle guard** (rework): self-cycle (`A extends A`) and mutual cycle (`A extends B, B extends A`) throw with the chain reported; diamond inheritance does not throw.

**Integration** (`js/test/idl-v2-parser-type-resolver.test.ts`):
- Two services each declaring `Packet` with different shapes → `resolveInService('A', ...)` vs `'B'` return their own definitions.
- Unknown service name → `undefined`.
- **Self-sufficient contract** (rework): a service resolver does NOT see program-level types — `resolveInService` returns `undefined` and `service.registry.hasType('Shared')` is `false`. Mirrors the JS test added in #1341.
- **Extends-chain visibility** (rework): an extender sees its base service's `types` through the transitive merge, both via `program.resolveInService('Child', ...)` and via `child.typeResolver.resolveNamed(...)`.
- **Sibling isolation** (rework): two services with same-named `Packet` types resolved through `SailsProgram.resolveInService` return distinct shapes — no cross-service bleed.
- `Envelope<[u8]>` → walker feeds struct fields through `substituteGenerics` + `resolveGenerics` to resolve `payload` from `T` to `[u8]`.

## Notes for reviewers

- `substituteGenerics` has a runtime cycle guard (visited set, throws with the full chain on detection) rather than just the JSDoc warning. Maps produced by `resolveGenerics` from parsed IDL can't produce cycles, but hand-merged or wire-sourced maps can.
- `SailsProgram.resolveInService` walks the AST directly (rather than going through `this.services`, which rebuilds `SailsService` + a fresh `TypeRegistry` on every access), and is backed by a lazy `Map<serviceName, Map<typeName, Type>>` that pre-merges each service's transitive extends scope. O(1) per call after the first invocation.
- `substituteGenerics` returns the same reference for subtrees that contained no substitutions — walkers traversing large trees with sparse type parameters don't allocate new nodes unnecessarily.
- The extends-chain merge in `_collectServiceScopeTypes` uses the same visited-set + immutable-copy cycle-guard pattern as `_substituteGenerics`, so a throw mid-recursion leaves no poisoned shared state.


js/src/sails-idl-v2.ts
```diff
@@ -215,11 +216,31 @@ export class SailsProgram {
...
+   */
+  resolveInService(serviceName: string, typeDecl: TypeDecl): Type | undefined {
+    if (typeof typeDecl === 'string' || typeDecl.kind !== 'named') return undefined;
+    const serviceUnit = this._doc.services?.find((s) => s.name === serviceName);
+    if (!serviceUnit) return undefined;
+    return (
+      serviceUnit.types?.find((t) => t.name === typeDecl.name) ??
+      this._doc.program?.types?.find((t) => t.name === typeDecl.name)
+    );
+  }
```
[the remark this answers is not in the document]
ukint-vs: Addressed in 8155c74a — SailsProgram now builds a lazy Map<serviceName, Map<typeName, Type>> index on the first resolveInService call and reuses it for O(1) lookups thereafter. Consumers that never touch this method pay nothing; walkers over large IDLs get constant-time resolution per node.

ukint-vs:
Rebased on master and addressed maintainer feedback:

1. **`TypeResolver` constructor reverted to `(types: Type[])`.** `[...ambientTypes, ...localTypes]` merging now happens at the call site in `SailsService`. The resolver no longer knows about the ambient-vs-local distinction; callers pass a single pre-merged array and rely on last-write-wins semantics for shadowing.
2. **Adapted to `TypeDecl::Generic` (from #1314).** `substituteGenerics` now targets the explicit `{kind:'generic'}` variant for type-parameter leaves; a bare `{kind:'named'}` decl is always a concrete user-type reference and is no longer used as a type-param fallback. `resolveNamed` returns `undefined` for `kind:'generic'`. Tests updated accordingly.

Cycle detection, chain resolution (`T → U → u32`), and same-reference returns on unchanged subtrees are preserved. 64/64 offline tests + lint green locally.

js/src/type-resolver-idl-v2.ts
```diff
@@ -21,18 +21,111 @@ export class TypeResolver {
...
+   * (`{ kind: 'generic' }`), and unknown names. Does not recurse into generics — callers
+   * that want the substituted inner shape should pair this with {@link substituteGenerics}.
+   *
+   * The returned `Type` is shared with the resolver's internal state — do not mutate it.
+   */
+  resolveNamed(type: TypeDecl): Type | undefined {
+    if (typeof type === 'string') return undefined;
+    if (type.kind !== 'named') return undefined;
+    return this._userTypes[type.name];
+  }
```
vobradovich, line 80: Consider changing the API to support both explicit lookup and TypeDecl ergonomics:

```ts
resolveNamed(type: TypeDecl): Type | undefined;
resolveNamed(name: string, generics?: TypeDecl[]): Type | undefined;
```

If generics are provided, I think `resolveNamed` should return the resolved concrete type rather than the raw generic definition. For example, this behavior feels wrong:

```ts
expect(resolver.resolveNamed(named('Envelope', ['u32']))).toBe(envelope);
```

Given `Envelope<T>`, resolving `Envelope<u32>` should return something shaped like:

```ts
{
  kind: 'struct',
  name: 'Envelope',
  fields: [
    { name: 'id', type: 'u32' },
    { name: 'payload', type: 'u32' },
  ],
}
```

Potentially omit `type_params` on the returned value because the result is concrete. That makes the API harder to misuse and avoids leaking generic-substitution details to callers.

js/src/type-resolver-idl-v2.ts
```diff
@@ -21,18 +21,111 @@ export class TypeResolver {
...
+    }
+    throw new Error('Unknown TypeDecl kind :: ' + JSON.stringify(type));
+  }
+
+  /**
+   * Build a substitution map from a user type's declared `type_params` and a concrete
+   * generics list (typically the `generics` field of a `{ kind: 'named', generics: [...] }`
+   * `TypeDecl`). Missing positions are omitted.
+   */
+  genericsSubstitutions(userType: Type, generics: TypeDecl[] = []): Record<string, TypeDecl> {
```
vobradovich: genericsSubstitutions feels like an implementation detail leaking out. Consumers should not need to know that a Type has type_params and manually zip them with a TypeDecl.generics array. That is resolver logic.

js/src/type-resolver-idl-v2.ts
```diff
@@ -21,18 +21,111 @@ export class TypeResolver {
...
+   * equivalent tree. Only `{ kind: 'generic', name: 'T' }` leaves whose `name` appears in
+   * `substitutions` are replaced; wrapper shapes (`Option<T>`, `Vec<T>`, `Result<T, E>`, custom
+   * generics, and any `named` decl) are preserved and their inner types substituted in place.
+   *
+   * The function recurses through replacement chains (`{ T: U, U: u32 }` resolves `T` to `u32`).
+   * Cyclic maps (`{ T: { kind: 'generic', name: 'T' } }` or `{ T: U, U: T }`) are detected at
+   * runtime and cause an error to be thrown rather than an unbounded recursion. Maps produced by
+   * {@link genericsSubstitutions} from a parsed IDL cannot create cycles.
+   */
+  substituteGenerics(type: TypeDecl, substitutions: Record<string, TypeDecl> = {}): TypeDecl {
```
vobradovich: `resolveGenerics`?

js/src/type-resolver-idl-v2.ts
ukint-vs: Addressed in b416c4f4 — `resolveNamed` now has two overloads:

```ts
resolveNamed(type: TypeDecl): Type | undefined;
resolveNamed(name: string, generics?: TypeDecl[]): Type | undefined;
```

When concrete generics are provided (via either overload or via `typeDecl.generics`), the result is a **concrete** substituted `Type` with `type_params` stripped. `resolveNamed(named('Envelope', ['u32']))` now returns exactly the shape you sketched:

```ts
{ kind: 'struct', name: 'Envelope', fields: [{id, 'u32'}, {payload, 'u32'}] }
```

Without generics, the raw user `Type` is returned unchanged for the sharing path.

js/src/type-resolver-idl-v2.ts
ukint-vs: Agreed — hidden in b416c4f4. `genericsSubstitutions` is now a private `_genericsSubstitutions` used only internally by `resolveNamed(name, generics)` and by the existing `getTypeDeclString` generic-wrapper registration. Consumers never need to zip `type_params` with a concrete generics list themselves.

js/src/type-resolver-idl-v2.ts
ukint-vs: Renamed to `resolveGenerics` in b416c4f4.

vobradovich (approved):

vobradovich:
## Design rationale: self-sufficient service IDL

The `ambientTypes` plumbing in `TypeResolver` / `SailsService` reflects a stricter scoping rule than what the parser permits today as an authoring convenience.
Recording the contract here so the runtime, the codegen, and `idl-gen` stay consistent over time.

### The rule

A **service IDL is self-sufficient and independently distributable.** A `<service>.idl` parses, validates, and resolves all of its `functions` and
`events` signatures using only its own `types {}` block (plus primitives and `Option` / `Result` / `NonZero*`).

Cross-scope type references are restricted to:

| From → To                               | Mechanism                                                           | Status                                                        |
| --------------------------------------- | ------------------------------------------------------------------- | ------------------------------------------------------------- |
| Service → its base service's types      | `extends Base@<interface_id>` (types merged into the extender)      | implemented                                                   |
| Program ctor → service type             | `Service::Type` qualified syntax                                    | TBD per `docs/idl-v2-spec.md` lines 419–420                   |
| Service → sibling service type          | **forbidden**                                                       | enforced by `rs/idl-parser-v2/src/post_process.rs::Validator` |
| Service → program type, **unqualified** | **soft-allowed by parser as authoring convenience**, NOT a contract | see below                                                     |

The parser pushes program types into the root scope before validating each service (`post_process.rs:32-44`). That's a DRY affordance for hand-written multi-service IDLs in a single file.
**Generators and the runtime must not rely on it for distributability.** A type used by a service `S` must be present in `S.types` for `S.idl` to be parseable on its own.

`program.types` should not be treated as an ambient scope for every service. The service IDL boundary needs to stay self-contained: a service’s functions/events/throws should be resolvable from that service’s own types plus any explicitly extended service interfaces, not from the enclosing program. extends is a service-to-service dependency keyed by interface_id, so it can be distributed independently from a concrete program. The namespace TBD in the spec is better interpreted in the opposite direction: program/root declarations may eventually reference service-local types via Service::Type, but services should not implicitly see program.types. Otherwise the same service IDL becomes decodable only when bundled with a particular program IDL, which breaks independent service distribution.

vobradovich (changes requested):
violates self-sufficient service IDL concept

vobradovich (approved):

PR #1352, opened 2026-05-13
feat: Add header-first Sails decode APIs

ukint-vs:
## Summary

This PR adds a header-first decoding path for the JS SDK and a small WASM custom-section reader for embedded Sails IDL. The goal is to let consumers decode calls, replies, errors, constructors, and events from raw message bytes using the v1 Sails header as the dispatch key, instead of requiring every caller to manually reconstruct route and entry metadata.

## Why

Sails payloads already carry enough routing information in the v1 header: interface id, route index, entry id, and header length/version data. Before this change, JS consumers had parsing and encoding helpers, but no single API that could take an incoming byte payload and resolve it back to the IDL entry that produced it.

That gap matters for indexers, explorers, tests, and app runtimes that observe replies/events after the fact. They need to decode messages from bytes first, then decide what entry they belong to. The new API makes that flow explicit and keeps malformed or unknown messages representable as structured `unknown` results instead of throwing for normal routing misses.

## Implementation Details

- Adds `SailsMessageHeader.tryReadBytes` support in the IDL v2 parser surface and uses it as the first step of JS dispatch decoding.
- Extends `SailsProgram` with a route/entry index built from the parsed IDL document, including service routes, constructors, command/query entries, events, return types, and throws types.
- Adds decode APIs for calls, replies, errors, constructors, and events. These return resolved entry metadata plus decoded values, or a typed `unknown` result for cases like bad header version, bad header length, missing service, missing entry, ambiguous route inference, decode failure, and trailing bytes.
- Preserves the existing `route_idx = 0` inference sentinel behavior. Expected-entry validation treats `0` as a wildcard on either side so route-inferred payloads are not rejected after header parsing.
- Adds `extractIdlFromWasm` for `sails:idl` custom sections. The section payload uses a tiny envelope with version and flags so raw and `deflate-raw` IDL text are both supported while remaining forward-compatible.
- Hardens the WASM section parser with typed parse/decode/UTF-8/size errors, decompressed size limiting, unsigned `u32` ULEB128 handling, and reduced unnecessary buffer copies.
- Adds shared JS/Rust fixture coverage for embedded IDL sections and dedicated decode tests/benchmarks for the dispatcher path.

## User Impact

Consumers can now construct a `SailsProgram` from IDL and decode observed bytes directly through stable high-level APIs. Unknown or malformed payloads become inspectable results, which is useful for explorers and indexers that should keep processing streams even when one message does not match the expected IDL.

Embedded IDL extraction gives tooling a local way to recover an interface from program bytes when a `sails:idl` custom section is present. This is intentionally separate from Rust/Gear build provenance metadata: standard WASM sections such as `producers` may exist before optimization, but optimized Gear artifacts can strip them.

## Validation

- `cargo test -p sails-idl-embed`
- `cargo test -p sails-idl-meta header`
- `pnpm --filter sails-js-types build`
- `pnpm --filter sails-js-util build`
- `pnpm --filter sails-js-parser-idl-v2 build`
- `pnpm --filter sails-js build`
- `pnpm --filter sails-js test:decode`
- `pnpm --filter sails-js test:bench`
- `cd js/parser-idl-v2 && pnpm jest --runInBand -- wasm-section.test.ts`
- `npx eslint -c eslint.config.mjs --ignore-pattern '.claude/**'`
- `git diff --check`


rs/idl-embed/src/lib.rs
```diff
@@ -222,6 +222,16 @@ mod tests {
         assert_eq!(extracted, Some(idl.to_string()));
     }
 
+    #[test]
+    fn extracts_shared_js_snapshot_fixture() {
```
vobradovich: Better remove this Rust test’s dependency on `js/test/fixtures/...`. It couples `rs/idl-embed` to the JS test fixture layout, so JS-only fixture moves or cleanup can break Rust tests. The Rust crate already has local embed/extract coverage; if we need a shared conformance fixture, it should live in a neutral cross-language fixtures directory rather than under `js/test`.

docs/idl-v2-spec.md
```diff
@@ -487,3 +487,20 @@ program DemoCanvas {
...
+Current codec tokens are `scale` and `ethabi`.
+The annotation describes which generated dispatch paths expose the item:
+
+- No `@codec` annotation means both dispatch paths are available where the target supports them.
+- `@codec: scale` means the item is available through SCALE/Gear dispatch.
+- `@codec: ethabi` means the item is available through Solidity ABI dispatch.
+- `@codec: scale,ethabi` means both paths are available.
+
+Solidity ABI dispatch is a separate ethexe/Solidity-facing path and is not implied by the Sails header alone.
+Tools that only implement the header-first Sails dispatcher must not silently decode `ethabi`-only entries as SCALE.
```
vobradovich: ## Codec Annotations

IDL items may carry `@codec` annotations to declare which codec-specific surfaces they participate in.

Current codec tokens are `scale` and `ethabi`.

- No `@codec` annotation means all supported codecs are available.
- `@codec: scale` means SCALE-capable tooling may expose the item.
- `@codec: ethabi` means Solidity ABI tooling may expose the item.
- `@codec: scale,ethabi` means both are available.

The annotation is declarative metadata for generators and tooling; it does not define a runtime dispatch protocol by itself.

vobradovich: Pls, move execution-specific parts to another doc, smth like `sails-message-dispatch.md`

js/types/src/decode.ts
```diff
@@ -0,0 +1,55 @@
...
+
+export type ResolvedEntry =
+  | {
+      kind: 'command' | 'query';
+      service: string;
+      fn: string;
+      route: string;
+      interfaceId: IInterfaceId;
+      entryId: number;
+      route_idx: number;
```
vobradovich:
```suggestion
      routeIdx: number;
```

vobradovich (approved):

